PR #4, opened 2015-04-25
Initial import

sagikazarmark:
This PR contains the initial import of interfaces and the following changes:
- Docblock cleanup (discussed with @egeloen in an earlier issue)
- Changed array notation to PHP 5.4 (It is shorter and does not mess with parenthesis)
- Since this is an interface package, the underlying interfaces should remain stable as well: psr/http-message is required explicitly

### DIFF
--- a/src/Message/InternalRequestInterface.php
+++ b/src/Message/InternalRequestInterface.php
@@ -19,18 +19,11 @@ interface InternalRequestInterface extends RequestInterface
     /**
      * Returns some data by name
      *
-     * @param string $name
+     * @param string|null $name
      *
      * @return mixed
      */
-    public function getData($name);
-
-    /**
-     * Returns the data
-     *
-     * @return array
-     */
-    public function getDatas();
+    public function getData($name = null);
 
     /**
      * Checks if the data exists
--- a/src/Message/InternalRequestInterface.php
+++ b/src/Message/InternalRequestInterface.php
@@ -21,7 +21,7 @@ interface InternalRequestInterface extends RequestInterface
      *
      * @param string $name
      *
-     * @return mixed
+     * @return string|array
      */
     public function getData($name);
 
--- a/src/Message/InternalRequestInterface.php
+++ b/src/Message/InternalRequestInterface.php
@@ -19,11 +19,18 @@ interface InternalRequestInterface extends RequestInterface
     /**
      * Returns some data by name
      *
-     * @param string|null $name
+     * @param string $name
      *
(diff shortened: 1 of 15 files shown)


src/Message/InternalRequestInterface.php
```diff
@@ -0,0 +1,140 @@
...
+ * @author GeLo <geloen.eric@gmail.com>
+ */
+interface InternalRequestInterface extends RequestInterface
+{
+    /**
+     * Returns some data by name
+     *
+     * @param string $name
+     *
+     * @return mixed
```
egeloen: `mixed` can be replaced by `string|array`

src/Message/MessageFactoryInterface.php
```diff
@@ -0,0 +1,108 @@
...
+     * @return boolean
+     */
+    public function hasBaseUri();
+
+    /**
+     * Sets the base URI
+     *
+     * @param null|string|UriInterface $baseUri
+     *
+     * @throws HttpAdapterException If the base uri is invalid.
```
egeloen: If it comes from the old package, I'm wrong... By reading the current implementation, this is an exception which comes from `phly/http` so it can't be an `HttpAdapterException`.

src/Message/MessageFactoryInterface.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public function setBaseUri($baseUri);
+
+    /**
+     * Creates a new request
+     *
+     * @param string|UriInterface                  $uri
+     * @param string                               $method
+     * @param string                               $protocolVersion
+     * @param array                                $headers
```
egeloen: `array` can be replaced by `string[]`. Same for internal request and response.

src/Message/MessageFactoryInterface.php
sagikazarmark: So what's the prognosis? Should it be simply removed?

src/PsrHttpAdapter.php
```diff
@@ -0,0 +1,50 @@
...
+    public function sendRequest(RequestInterface $request);
+
+    /**
+     * Sends PSR requests
+     *
+     * @param RequestInterface[] $requests
+     *
+     * @return ResponseInterface[]
+     *
+     * @throws MultiHttpAdapterException If an error occurred when you don't provide the error callable.
```
egeloen: If an error occured is enough since we can't pass a callable anymore.

PR #7, opened 2015-05-04
Refine PHPDoc

egeloen:
This PR refines PHPDoc, for example if an uri is not valid, it will throw an `\InvalidArgumentException` comming from `phly/http` and in other cases can come from us when I will update the `php-http/adapter-core`. Additionally, the data can accept a `StreamInterface` if no files are provided.

### DIFF
--- a/src/HttpAdapter.php
+++ b/src/HttpAdapter.php
@@ -11,6 +11,7 @@
 
 namespace Http\Adapter;
 
+use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\UriInterface;
 use Psr\Http\Message\ResponseInterface;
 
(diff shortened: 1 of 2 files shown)


src/PsrHttpAdapter.php
```diff
@@ -37,7 +37,8 @@ public function sendRequest(RequestInterface $request);
      *
      * @return ResponseInterface[]
      *
-     * @throws MultiHttpAdapterException If an error occurred.
+     * @throws \InvalidArgumentException If an request is not valid.
```
sagikazarmark: "A" request

src/HttpAdapter.php
```diff
@@ -128,13 +137,14 @@ public function options($uri, array $headers = [], $data = [], array $files = []
...
-     * @param array|string        $data
-     * @param array               $files
+     * @param string                       $method
+     * @param string|UriInterface          $uri
+     * @param string[]                     $headers
+     * @param array|string|StreamInterface $data
+     * @param array                        $files
      *
-     * @throws HttpAdapterException If an error occurred.
+     * @throws \InvalidArgumentException If an argument is not valid.
```
sagikazarmark: "If an instance of StreamInterface and $files are passed together"?

egeloen:
Dropped comments related to exceptions since there are lot of cases where these exceptions can be thrown.

sagikazarmark:
:+1:

Issue #10, opened 2015-05-15
Create common Exception interface, break exceptions into smaller parts


sagikazarmark:
It is a common practice to introduce a generic Exception interface so that all exceptions thrown by this library can be caught in a single catch.

Also, as discussed on gitter, we should create more exceptions instead of one or two big ones.

ddeboer:
:+1:

At least differentiate between client exceptions (400s), server exceptions (500s) and networking errors (timeouts and such).

sagikazarmark:
That's a good point.

Currently the separation is based on single and multi requests.

Can you open a separate issue for that?

PR #15, opened 2015-05-16
Exceptions

sagikazarmark:
I decided to remove exception from the package entirely and replace them with interfaces. This also means we have to move them to the core. Since the core is required by EVERYTHING, I don't think it is a problem.

I also introduced a top Exception interface which should be implemented by all exceptions in the adapter.

I also removed all test cases since in this package only the exceptions were tested.

This PR solves #10


sagikazarmark:
See #24

PR #23, opened 2015-05-23
Depend on stable psr/http-message

ddeboer:


sagikazarmark:
Could you please make it `~1.0`. It is just easier to read.

ddeboer:
Updated.

sagikazarmark:
Thanks

Issue #38, opened 2015-08-10
Terminology


sagikazarmark:
I am thinking about the following:

We could make adapter/client a set of interfaces which are recommended for client implementations. This way if an HTTP Client PSR is created once, we already have something.

The adapter naming is useful in the context of creating adapters wrapping clients. But for the long term, clients themselves should have a common interface which makes the adapter naming a bit vague.

I remember @jeromegamez mentioned something similar.

What do you think?

hannesvdvreken:
Not sure, I believe the interfaces `Http\Adapter\HttpAdapter` and `Http\Client\HttpClient` are super clear and minimal (does not violate interface segregation principle at all!).

What different interface would you suggest?

sagikazarmark:
Actually, I would like to preserve the two separated interfaces. They are fine. My particular idea is to rename the adapter to something which just could be used as a client itself. For example Guzzle 7 could implement what we call Adapter interface (and the Client interface as well). From this point, the adapter is not an adapter anymore, but the client itself.

So maybe we could call these things `PsrHttpClient` and `HttpClient`. Actual adapters themselves can still be called adapters. But the base interface should not suggest if it is implemented by an adapter or an actual HTTP Client.

hannesvdvreken:
Good idea! Although I don't know if adoption is high enough and whether enough http client maintainers know about this interface.

sagikazarmark:
The point is that we make it possible. Also, if this interface gets to a PSR once, we don't have to change our entire architecture....better do it now.

sagikazarmark:
The two interfaces are good enough IMO. We just need proper naming. Also, I am thinking about to put the interfaces into one package. This also solves the exception problem I am struggling with: currently the adapter and the client interface have separate exceptions which does not make much sense IMO. They serve the same purpose.

hannesvdvreken:
I agree. Also on the same package for both interfaces. And where would the client implementation go?

sagikazarmark:
Which client implementation? [This one](https://github.com/php-http/adapter-client)?

Currently I only have an imaginary state in my mind (as I explained it here), but I don't really have an exact implementation/naming strategy (yet).

@ddeboer can you also join the discussion? You've already used the adapter, I am curious about your points on this.

hannesvdvreken:
Yes. php-http/adapter, php-http/adapter-client and php-http/client could be one repository imho. Btw I'm doing a talk at Laracon EU in 2 weeks where I will mention php-http.

[...]

Issue #26, opened 2015-05-25
Separate interfaces for client, server and networking exceptions


ddeboer:
As discussed in #10.

sagikazarmark:
I have to mention that it is not that easy to differentiate between errors, since each HTTP Clients handle errors differently: some throws exceptions for 4xx status code responses, some don't. Currently we return every response if any, even if 4xx or 5xx responses.

ddeboer:
That’s an option too, at least for requests that return a response. I think you need to throw an exception at least for cases where a networking timeout occurs.

sagikazarmark:
That's true, AFAIK we do it ATM.

PR #40, opened 2015-08-23
Terminology

sagikazarmark:
This PR tries to solve #38

Some points to discuss:
- [x] Exceptions in the package: concrete implementations? Exception naming?
- [x] Interface names: PsrClient, HttpClient, Client? (proposed by @dbu HttpPsrClient, HttpMethodsClient, HttpClient)
- [x] Add ClientTemplate: better name?

To do before merging:
- ~~Squash commits!!!!!!!~~

To do after merging:
- [ ] Move the package to client (preserve the package here as well for historical reasons (0.1 release)?)


sagikazarmark:
@hannesvdvreken FYI

hannesvdvreken:
@sagikazarmark I'm watching this repo so saw the PR ;-) I'll have a look tomorrow.

sagikazarmark:
Well, my personal experience is that people tend to look at Github messages more actively if they are specifically mentioned. :wink: 

Sure, no hurry.

hannesvdvreken:
I've gone through the changes quickly and I think it's very clear now.

The adapter packages just implement the PsrHttpClient interface, right? There will be a HttpClient and a Client class which wraps a PsrHttpClient and provides the interfaces and still expose the PsrHttpClient methods.

Didn't take an in depth look to find mistakes because time constraints, but I thrust it will be close to perfect ;-)

sagikazarmark:
> The adapter packages just implement the PsrHttpClient interface, right?

Basically, yes.

> There will be a HttpClient and a Client class which wraps a PsrHttpClient and provides the interfaces and still expose the PsrHttpClient methods.

The plan is to have something called adapter client which is a wrapper around a PsrHttpClient and yes, still implements it.

Something like:

``` php

class AdapterClient implements Client
{
public function __construct(PsrHttpClient $psrHttpClient) {}
}
```

sagikazarmark:
@hannesvdvreken thanks for feedback. What do you think about the TODOs I mentioned in the issue description?

src/HttpClient.php
```diff
@@ -0,0 +1,145 @@
+<?php
+
+namespace Http\Client;
+
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\UriInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Interface for HTTP conventional methods
```
dbu: maybe say "Interface for HTTP requests without the PSR-7 RequestInterface"

sagikazarmark: Hm. Sounds good, but does it really explain what this interface is about? The point is rather about HTTP methods than the lack of RequestInterface.

src/Client.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Http\Client;
+
+/**
+ * Interface for both HTTP Client type
+ *
+ * @author Márk Sági-Kazár mark.sagikazar@gmail.com>
+ */
+interface Client extends PsrHttpClient, HttpClient
```
dbu: so implementations would either unwrap the request and call the underlying client, or the methods of HttpClient would be factory methods to generate the request?

that looks convenient to me. the documentation should state clearly that you should specify PsrHttpClient or HttpClient rather than the combined interface unless you really need both.

about naming: i think we should call Client HttpClient. leaves the question what the thing now called HttpClient should be called? LegacyHttpClient sounds too negative. SimpleHttpClient is a bit weird. its actually a RequestFactoryHttpClient but that name is a bit long.

sagikazarmark: Well, this interface should rather be implemented by actual HTTP Clients, like guzzle. Also eases implementing both interfaces extended by this one. Probably this is only for type hint convenience. The naming is on the TODO list, because I am not sure about the current state either. The reason why I named HttpClient this way is that it is clearly about HTTP methods. Probably the word "method" could be used somehow in the interface? I am open for ideas.

BTW I agree about the documentation should clear all those questions that originates from naming.

src/HttpClient.php
dbu: then maybe just add a second line saying: "Use this interface when you do not have PSR-7 RequestInterface instances available."

sagikazarmark: :+1:

src/Client.php
dbu: maybe HttpPsrClient and HttpMethodsClient?

sagikazarmark: Sounds good.

hannesvdvreken: > the documentation should state clearly that you should specify PsrHttpClient or HttpClient rather than the combined interface unless you really need both.

Agreed

sagikazarmark: Actually I would close from the other way as well: if you need both, you SHOULD implement Client instead. It eases type-hinting for both interfaces.

dbu: another thing: adapter could provide two traits for client
implementations to provide the other half. one that has the request
factory and the http methods and calls send(), another one that looks at
the request and calls the appropriate http methods.

then the two sub interfaces would never really be used alone and all
clients could implement the main interface without much effort.

sagikazarmark: Partly agree: we have something called ClientTemplate which implements all HTTP method using send. However I am not sure about the other. When the two interfaces are implemented together, `send` should use a message factory to create a request object then use `sendRequest` to actually send the request.

sagikazarmark:
Another major change I am thinking about: currently post, put, patch, delete and options request in HttpClient have two type of body fields: `$data = [], array $files = []`.

`$data` can be string, array, StreamInterface, while `$files` are always an array of files.

Furthermore, if `$data` is stream interface than `$files` should not be passed (hence the body is already prepared).

I see the reasoning behind, it was hard for @egeloen to create an easy, but full solution. It is still a little bit weird, so I am thinking about the following:

I would create a Body interface with the following implementations:
- Stream: accepts a stream interface
- String: accepts a string body
- Data: accepts an array body
- Files: accepts files
- Combined: accepts a Data and a Files object

With the exception of Stream these objects would be string castable which can be used then to create a Stream object.

This is obviously extra complexity, but I came up with the idea, because I think it can be confusing this way as well. Also, much of the body creating logic can be moved to the Body classes and the client should only handle string to stream conversion.

What do you think?

sagikazarmark:
Let's continue the body discussion in #42

src/Client.php
dbu: i think traits are more suitable than base classes for this.

regareing body: these would be value objects, no factory? i kind of wonder
if its not the same problem as the request factory has to solve. maybe drop
the methods in favor of the factory?

sagikazarmark: ClientTemplate is a trait of course.

Yes, bodies would be value objects. Drop what methods?

You can take a look at the body branch. Currently the interface accepts a plain string body, a StreamInterface or a Body object. Body objects must be string castable. Also, in some cases (for example in case of Files) the request headers must be altered, so I am thinking about to add an alterRequest method to the Body interface which allows to add required headers.

dbu: maybe i miss something, but all the methods except sendRequest seem like a repetition of the request factory to me. the client would be much simpler if it only had that one method and would not need to bother to be a factory as well.

especially with the topic of the body value objects - sounds like something for the request factory. the RequestInterface already handles all the stream and whatnot stuff. if the request factory is more useful with body value objects that is fine, but the client should not care imho.

sagikazarmark: I'm not sure we are on the same track. The HTTP method interface contains all the conventional HTTP methods for convenience purposes. Of course it is an optional interface, but in some cases it is easier to do `$client->get('/path');` than using a message factory (this is what you call a request factory I guess) and creating requests that way. This is "for the public". An API client will probably construct the requests internally, while in a spaghetti scenario you probably don't want to do that.

IMO the message factory SHOULD NOT care about what the input really is (like is it a multipart form or not?). The message factory and the PsrHttpClient interface assumes that you have prepared data for input. In case of calling a simple get method this is not the case. This is why there were `$data` and `$files` parameters.

With the change I proposed in #42 the special case, when you want to send files in a POST request is still possible, but requires extra efforts from the user. If you want to send string or a StreamInterface, just pass the `$body` parameter to the `post` method. If you want files and other extra sauce for your spaghetti, you have to do a little more than that. But still possible.

Currently I implemented a few Body variations in the body branch, but I am thinking about removing them and leaving only the interface in the adapter/client package. Body implementations can go into a separate package, if needed.

sagikazarmark: Also, please continue this discussion in #42.

dbu: added some comments in #42. now i got what you want to do and it does make sense. and sorry for calling message factory "request factory". i meant indeed message factory.

the discussion about what methods the HttpClient needs should go here so i continue that here :-)
i see the value in having a convenient simple get/post/put method signature. but i would only do that with simply string body and once you need file upload or stream bodies, go to the request factory. although technically this is just syntactic sugar around the message factory - so i guess we can also support the Body interface here to make things more consistent.

ClientTemplate seems a strange name for that trait. can we find a better name? ok, not in this discussion probably.

sagikazarmark: The method signature in the body branch changed to: `string|StreamInterface|Body`. I guess it is simple enough, still allows advanced usage.

sagikazarmark:
@dbu Ideas for ClientTemplate naming? HttpMethods for example? Since it only "implements" the HTTP methods, `send` still needs to be implemented. Not sure though, because it might collide with HttpMethodsClient.

dbu:
HttpMethodsClient as interface and HttpMethods as trait sounds good to me.

sagikazarmark:
#26 should also be addressed here.

src/HttpMethodsClient.php
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param string|UriInterface $uri
+     * @param array               $headers
+     * @param array               $options
+     *
+     * @throws Exception
+     *
+     * @return ResponseInterface
+     */
+    public function get($uri, array $headers = [], array $options = []);
```
dbu, line 29: are you aware that technically a GET request can also have a body? elasticsearch is doing this extensively for example.

not sure if we need to care here or just point people to the message factory for that use case. on the other hand having the body would make the method signatures consistent

sagikazarmark, line 29: Hm, I wasn't aware of that. What about HEAD and TRACE?

I would say this is a rare use case and can be achieved by using `send`.

dbu, line 29: HEAD is just a different kind of GET. but yeah, lets not cover these
cases here.

dbu:
do we need to resolve the BatchResponse #47 before merging this?

sagikazarmark:
I don't think we need to. But we should before putting it into review phase. I am already working on it.

ddeboer:
Sorry to chime in so late, but I have some doubts about the mix between interfaces and ‘convenience’ classes such as those in the `Body\` namespace, as it seems to break single responsibility. An adapter package (such as php-http/guzzle6-adapter) would only need the interfaces, as the body etc. classes are already available in Guzzle itself. Or is the idea that the adapters also translate between Guzzle’s own Body etc. implementation and the PHP-HTTP ones?

sagikazarmark:
The Body classes are supposed to replace the InternalRequest object in Ivory. The point is that you can use these clases to provide custom type data (like files) to the convenience method calls. They are completely optional and can be translated into an array of content headers and a string which then can be passed to the message factory. Guzzle 6 adapter probably won't need it, as they are related to the convenience methods, adapters does not need to implement those. I plan to create a separate "adapter/client" for that which accepts an HttpPsrClient and implements HttpMethodsClient.

ddeboer:
That they are optional is exactly the problem that I have with them. The Body classes have mostly to do with messages, don't they? In that case, they could go into a separate message package that only clients that need them depend on.

sagikazarmark:
The Body interface is part of the contract, it is used in HttpMethodsClient. The basic implementations provided are small. That said, it is still implementation and you might be right about moving them into a utility package. There is another BatchRequest trait work in progress which is also something like that.

My only concern about it is in that case the utility package must require the contract, which in turn can lead to a dependency hell: an adapter/client requires both the contract and the utility. If versions doesn't match, you can't install the adapter.

dbu:
i think the utility repo becomes more and more of a thing. everything
not part of the contract should not be in this repo.

dependency is a thing we need to be careful about anyways. we can't
change anything in an interface in the contract without bumping the
major version. which is something we should be very careful to avoid doing.

the utility is not making the problem worse. it will just have a minimum
requirement of the major version of the client, and every adapter should
trust both utils and contract to be proper semver, that is accept any
minor version that is new enough for it.

even if we bump the contract, the utilities might not need to bump their
major version if what they do is not affected.

sagikazarmark:
I hope that major version bumps will be extremely rare, so we might not even face this dependency hell issue we are afraid from. The only condition is to have well-designed interfaces, which we are working on right now.

joelwurtz:
IMO we can leave utility classes in this package (client), as if they are not used they don't add anything expect for the disk size.

sagikazarmark:
Yes, they do. Bandwidth, package complexity, autoloading. These are minor things, but imagine the package installed a thousand times.

sagikazarmark:
Any last words before merging this one?

sagikazarmark:
Let's continue the utility discussion in #56

joelwurtz:
I think this can be merged, anyways it's an alpha release and there is still room for other alpha release if big change come up with new discussions

joelwurtz:
A remark about the Body, if it is split into an Utility package, all adapter implementing the HttpMethodsClient interface will then have to depend on this package ?

sagikazarmark:
Nope. The Body interface remains in this package. Optionally the implementations should be moved to the utility package.

src/Body.php
```diff
@@ -0,0 +1,33 @@
...
+     * Content-Length is calculated automatically if possible
+     *
+     * @return array
+     */
+    public function getContentHeaders();
+
+    /**
+     * Convert data to a format which can be used to create a proper PSR-7 Stream
+     *
+     * @return string|StreamInterface
```
joelwurtz, line 28: PSR7 Only allow StreamInterface no string

sagikazarmark, line 28: Nope, string is a valid return type here as well. Keep in mind that this interface is used in the MethodsClient where you probably use a MessageFactory.

joelwurtz, line 28: Yes but it may conflict with some implementations like diactoros where the string is an url to stream and not a content. But maybe the string here as the same purpose ? In this case we should add documentation about the string return value.

sagikazarmark, line 28: We have a custom MessageFactory interface which is recommended to be used. This ensures you can use whatever PSR7 implementation in your clients. There is already some kind of documentation about it, for reference check the docs repo.

Presonally I think that the diactoros behavior is a bit weird. I think using string as a simple body is much more logical than using it as a URL to a stream.

joelwurtz, line 28: I agree, i'm working on the socket adapter right now and using diactoros (only for the init puprose) feel weirds when implementing the send method :/

sagikazarmark, line 28: Check the docs here: http://php-http.readthedocs.org/en/latest/

We have a zero-config discovery service which let you just use the available implementation AND message factory.

dbu, line 28: nonetheless we should explain what the string we allow to return here is. if we allow a string at all - it seems like an ambiguity. ideally we would even eliminate string as return completely. otherwise a consumer needs to check whether it got a string or a stream and do different things...

src/Body.php
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @throws Exception
```
dbu, line 30: this being an interface we could also say what circumstances might typically lead to exceptions.

src/Body.php
sagikazarmark, line 28: We have already discussed this in another issue. Body objects would need a message factory to return a stream object. And that's one simple IF to check the input parameter.

The alternative is to remove body types entirely and use "generators" which can populate the request based on a specific input. It would certainly make the contract cleaner. And we could move that code to the utility package.

sagikazarmark, line 28: That would also mean, that requests sent through HttpMethodsClient must be populated from the generator manually.

PR #42, opened 2015-08-25
Add Body implementations

sagikazarmark:


hannesvdvreken:
At Laracon EU right now. Best to not expect much from me till Thursday ;-)

sagikazarmark:
No problem. A gitter conversation after that?

hannesvdvreken:
Yup

dbu:
i think this indeed looks like a good idea. in this case we won't have a body factory but just ask users to instantiate the appropriate object. i think for that reason it makes sense to have the common use cases right here in the contract package.

the MessageFactory should also know about the body interface i think and use it when passed a body. this does not need to be a hard dependency as `instanceof` can check for a non-existing class without triggering an error...

sagikazarmark:
> i think for that reason it makes sense to have the common use cases right here in the contract package.

Well, I am struggling with that a little. But creating a php-http/body package for just a few classes which are actually part of the core doesn't seem to make much sense.

dbu:
i would say dont exagerate with granularity :-) as long as those body impl
do not add new dependencies its fine to have them directly.

sagikazarmark:
> the MessageFactory should also know about the body interface

Actually the MessageFactory has nothing to do with the Client/Adapter thing. The MessageFactory package has been built for PSR-7 and I would like to make it as small as possible.

However I see the reason behind why it would make sense to handle Body classes in the message factory: some content might need custom headers. To solve this I am going to add a `getContentHeaders` method to the `Body` interface.

sagikazarmark:
I tried to find better names for body implementations. Opinions?

sagikazarmark:
Merging this anyway, naming can be solved as part of Terminology.

src/Body.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Http\Client;
+
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Allows a more input types for body data
```
dbu, line 8: grammatically confusing

src/Body.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
+interface Body
+{
+    /**
+     * Returns a set of headers which is needed to correctly send the body
+     *
+     * Note: these headers SHOULD be overwritten if any matching header is passed manually
```
dbu, line 17: should be overwritten by what? from a user perspective of this library, this could be confusing

src/Body/Files.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+class Files extends Multipart
+{
+    /**
+     * @var array
+     */
+    protected $files;
+
+    /**
+     * @param array       $files
```
dbu, line 20: maybe say "absolute paths to files"?

src/Body/PlainText.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->string = $string;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getContentHeaders()
+    {
+        return [
+            'Content-Type' => 'text/plain',
```
dbu, line 33: why do we need a special body type for this? should we not simply pass the content type header and the body as a string in this case? if a body class helps a lot, the content type should not be hardcoded but we should have a TextBody or StringBody something like that with a second constructor argument for the Content-Type

src/Body.php
sagikazarmark, line 8: Ah, probably mixed two different wording. "Allows special input as body"?

src/Body.php
sagikazarmark, line 17: Well, yes, probably this notice should not be here. The point is that headers returned from a body should be overwritten by the headers manually passed to the client. Just to allow full control for the user.

src/Body/Files.php
sagikazarmark, line 20: Good point

src/Body/PlainText.php
sagikazarmark, line 33: You are right. This body type was born at the time I wanted to typehint for the body interface in the HttpMethodsClient. Since then I just worked on it, but didn't rethink if it is useful or not.

src/Body.php
dbu, line 8: :+1:

src/Body.php
sagikazarmark, line 17: "Note: these headers get overwritten by headers manually passed to the client"?

dbu, line 17: maybe: "Note: When creating the HTTP message, the factory SHOULD overwrite matching headers returned by this method with those that are passed manually." ? and could we change that to "factory MUST" ?

dbu, line 17: sounds good too, yes

Issue #41, opened 2015-08-25
Make MultiException final


sagikazarmark:
I can't really imagine a case where this type of exception should be extended.

PR #43, opened 2015-08-26
Refactor exceptions

sagikazarmark:
This PR tries to address the following:
- Separated exceptions (#26)
- Multi/Batch exception is final (#41)
- Better naming
- Exceptions are immutable (removed all setter/cleaner methods)

Feel free to start discussion about any of the above.

...let me start:

Setters are completely removed from exceptions (no `setResponse`, `setRequest`). While I think this is the better way, I can imagine a case when some kind of event driven logic wants to modify the exceptions. Any ideas? Should we preserve the setter logic or force any event-driven one to create a new exception?

Differences from Guzzle: Guzzle has a Request-Response exception pair, plus almost everything extends request exception. One difference is that the Network exception does NOT extend the request exception. It can be used in cases where the request does not get to a state where it makes sense to return it. Like a connection timed out. Another difference is that Response exception is called Http.


sagikazarmark:
I am waiting for feedback from @hannesvdvreken @dbu

src/Exception/BatchException.php
```diff
@@ -0,0 +1,110 @@
...
+    private $exceptions;
+
+    /**
+     * @var ResponseInterface[]
+     */
+    private $responses;
+
+    /**
+     * @param TransferException[] $exceptions
+     * @param ResponseInterface[]   $responses
```
dbu: indention is wrong

src/Exception/BatchException.php
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @return boolean
+     */
+    public function hasExceptions()
+    {
+        return !empty($this->exceptions);
+    }
+
+    /**
+     * Returns all responses
```
dbu: we should explain what this is. is it all responses of the transaction? or only those that have an error? is there a way to associate the individual response with an exception and vice-versa?

src/Exception/HttpException.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+namespace Http\Client\Exception;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Thrown when both a request and a response are available
```
dbu: or maybe "Thrown when a response was received but has an error status code. This exception always provides the request and response objects."

src/Exception/RequestException.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Http\Client\Exception;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Thrown when a request is available
```
dbu: "Base exception for when a request failed. This can be because of network problems or the status code of the response."

src/Exception/RequestException.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
+class RequestException extends TransferException
+{
+    /**
+     * @var RequestInterface
+     */
+    protected $request;
```
dbu: why does this need to be protected? the request should be injected in the constructor and we have a getter method here. private should work.

src/Exception/TransferException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Http\Client\Exception;
+
+/**
+ * Thrown when an error related to sending/receiving the request/response occurs
```
dbu: "Base exception for all php-http exceptions."

basically its either RequestException or BatchException.

dbu:
looks good! i added some comments on the phpdoc and one visibility issue i suspect.

it will be tricky to ensure that all adapters actually catch all relevant exceptions and identify them to throw the right exception in turn.

src/Exception/ClientException.php
```diff
@@ -0,0 +1,13 @@
...
+namespace Http\Client\Exception;
+
+/**
+ * Thrown when a client error (4xx) is encountered
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
+class ClientException extends HttpException
+{
+
```
hannesvdvreken: empty line is not psr-2, sorry

src/Exception/HttpException.php
```diff
@@ -0,0 +1,84 @@
...
+     */
+    public function __construct(
+        $message,
+        RequestInterface $request,
+        ResponseInterface $response,
+        \Exception $previous = null
+    ) {
+        $this->response = $response;
+
+        parent::__construct($message, $request, $previous);
```
hannesvdvreken: I'd put the parent call first or last.

src/Exception/HttpException.php
```diff
@@ -0,0 +1,84 @@
...
+     * @param ResponseInterface $response
+     * @param \Exception|null   $previous
+     *
+     * @return HttpException
+     */
+    public static function create(RequestInterface $request, ResponseInterface $response, \Exception $previous = null)
+    {
+        $code = floor($response->getStatusCode() / 100);
+
+        if ($code == '4') {
```
hannesvdvreken: I'd use strict comparison and int 4 & 5.

dbu: sad that the response does not have the isClientError() / isServerError() methods of the symfony response object :-(

i would not go and divide but simply
$code = $response->getStatusCode();
if ($code >= 400 && $code < 500)
and so on. more readable and probably more efficient.

hannesvdvreken:
:+1:

For documentation: here's the inheritance tree for Exceptions:

```
\InvalidArgumentException
`- \Http\Client\Exception\InvalidArgumentException (implements \Http\Client\Exception)

\RuntimeException
`- \Http\Client\Exception\RuntimeException (implements \Http\Client\Exception)
   `- \Http\Client\Exception\TransferException
      |- \Http\Client\Exception\BatchException
      `- \Http\Client\Exception\RequestException
         |- \Http\Client\Exception\NetworkException
         `- \Http\Client\Exception\HttpException
            |- \Http\Client\Exception\ServerException
            `- \Http\Client\Exception\ClientException
```

Now copied here: https://github.com/php-http/documentation/issues/7 to be incorporated in documentation.

dbu:
not sure if we want to put that into the doc. there is one mistake here, ServerException extends HttpException, not RequestException. (the code is correct, but the tree in above post not)

hannesvdvreken:
@dbu You're right. Edited the comment.

src/Exception/BatchException.php
sagikazarmark: This is actually something that I ported from Ivory and Guzzle. In Ivory the array key (order of elements) identifies exception-response pairs. This method returns only the errored messages.

src/Exception/ClientException.php
sagikazarmark: Good to know, I always wondered if there is a standard for empty classes.

src/Exception/RequestException.php
sagikazarmark: Well, we provide an extension of this class, where you might need to access this variable. But you are right, this should probably be private.

sagikazarmark:
Hey guys,

Thanks for all the comments, I really appreciate it.

@dbu thanks for helping out with descriptions, I am not really good at it.

About ensuring that the correct exceptions are thrown: the interface declares that exceptions must implement out interface, so even if the exception type is not correctly determined, it is still possible to catch all exceptions thrown by the client.

I wonder if we should be strict about private types. Actually, I am not against it. We could even make Client and Server exceptions final, since those are thrown by the HttpException itself.

@hannesvdvreken Thanks for the inheritance tree, looks good. We should definitely put it in the docs once we get there (or at least something similar. I admit that the inheritance tree is a bit complex and it is needed to be coved in the docs).

src/Exception/BatchException.php
dbu: lets explain the semantics in the phpdoc to make it more useful and accessible for users. the doc says "Returns all responses" but if i understand you correctly, it should say "Returns all failed responses. The array keys are the same as in getExceptions". and we should also update the getExceptions phpdoc to say it has the same keys as getResponses. (though i expect most of those exceptions to carry the response in them again, no?). which brings me to wonder why we need this method at all? if there is an exception before we get a response, there is no response obviously. if we raise the exception because of a 4x or 5x response, we can put the response into that exception. what is the use case for having the responses separately?

dbu:
yes, the doc should definitely explain the exception concept. the tree is imho nice but the most important is explaining in what situation which exception will be thrown. is there an issue about for doc already where we can add this?

sagikazarmark:
Not yet. The main reason I neglected documentation so far is the domain: @egeloen owns it and we cannot use it.

dbu:
https://github.com/php-http/documentation/issues/7

sagikazarmark:
Network, Client and Server exceptions are `final`. Although I am not sure about the Network exception. Client and Server can be used in limited number of cases (namely 4xx and 5xx errors). Network exception might be used in cases like timeout, connection refused. While I don't think we should include those in the contract package, should we allow extending the NetworkException in such cases?

dbu:
i am not that worried with people extending that exception so lets not
make the network one final.

src/Exception/BatchException.php
sagikazarmark: I agree, there is unnecessary complexity here. This thing is ported as is from Ivory. However this exception also handles other exceptions which does not even have a response. So I think we should just stick to exceptions. If a response is available, it should be carried in the exception, as you said.

sagikazarmark: Another possibility is to store exceptions in an SplObjectStorage indexed by the request objects. This would be the best way to identify every exceptions. I am also thinking about adding a BatchResponse class as well, which also holds an SplObjectStorage internally, which would contain responses indexed by requests. WDYT?

dbu: that sounds like a very interesting and useful idea. though for an alpha1 just having the exceptions would already be something. 

we can still add SplObjectStorage separately. if we go with SplObjectStorage we should warn users about http://php.net/manual/de/class.splobjectstorage.php#114059 however (in foreach, the iterator value is the spl object "key" object and the iterator key is meaningless)

sagikazarmark: Well, I would keep SplObjectStorage an internal thing. Also, about the batch response: I more and more have the feeling that it is necessary. Currently it is impossible to pair requests/responses if an exception is thrown, because array indexes are probably messed up that way.

I am also thinking about if exceptions in case of a batch request should be thrown at all, or just returned in a general response object. Using this Batch exception, successful responses are unable to be retrieved.

sagikazarmark: But I think that should be resolved in a separate issue.

dbu: lets open a new issue for that. and check what guzzle and the others do
for this. (or even curl php).

and lets drop the getResponses from here as that definitely does not belong

sagikazarmark: See #47. I've already dropped the getResponses. Checking guzzle right now...

dbu: you still need to push dropping getResponses. though if we decide to drop BatchException we don't need this whole class.

PR #48, opened 2015-09-21
Add BatchResult

### LINKED ISSUE #47: Batch exceptions/responses
With the terminology branch merged, I only see a few questions we need to clear, this is one of them.

Currently parallel requests work like this: if the client supports it, send requests parallel, if not, emulate it by sending them one by one. In the latter case exception have to be caught and stored for later processing.

The return value of the parallel request function is currently an array of Response objects. The main problem with that: it is hard to identify which response belongs to which request. In an ideal case, array indicies match. The proposal here is to return an object which can return a response for a specific request. Internally we could use SplObjectStorage to achieve that.

In exceptional cases, currently a BatchException is used. The problem is the same as earlier: we cannot pair requests to exceptions. SplObjectStorage could help in this as well.

Another limitation is that currently only TransferExceptions are allowed to be stored in the batch exception. However it is possible that a request is rejected because of some sort of incompatibility: invalid method, etc. In these cases the underlying client might throw an exception which cannot be transformed to a TransferException, rather an InvalidArgumentException. So we either allow any exceptions to be added for a request (which implement our interface of course) or we require to convert that exception to a TransferException?

And a third problem: in case of any exceptions thrown, all the requests should still be processed. However, if we throw a batch exception, we won't get the successful requests. A solution for this: remove batch exception and use a combination of batch exception and response proposed earlier. This way we get all the results for requests (either response or exception) and don't block the processing.

sagikazarmark:
Fixes #47

### DIFF
--- a/src/BatchResult.php
+++ b/src/BatchResult.php
@@ -2,7 +2,7 @@
 
 namespace Http\Client;
 
-use Http\Client\Exception\InvalidArgumentException;
+use Http\Client\Exception\UnexpectedValueException;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
@@ -46,15 +46,15 @@ final class BatchResult
      *
      * @return ResponseInterface
      *
-     * @throws InvalidArgumentException
+     * @throws UnexpectedValueException
      */
     public function getResponseFor(RequestInterface $request)
     {
-        if (!$this->responses->contains($request)) {
-            throw new InvalidArgumentException('No response can be found for the given request');
+        try {
+            return $this->responses[$request];
+        } catch (\UnexpectedValueException $e) {
+            throw new UnexpectedValueException('Request not found', $e->getCode(), $e);
         }
-
-        return $this->responses[$request];
     }
 
     /**
--- a/src/BatchResult.php
+++ b/src/BatchResult.php
@@ -17,39 +17,9 @@ final class BatchResult
      */
     private $responses;
 
-    /**
-     * @var \SplObjectStorage
-     */
-    private $exceptions;
-
     public function __construct()
     {
         $this->responses = new \SplObjectStorage();
-        $this->exceptions = new \SplObjectStorage();
-    }
-
-    /**
-     * Checks if a request is successful
-     *
-     * @param RequestInterface $request
-     *
-     * @return boolean
-     */
-    public function isSuccessful(RequestInterface $request)
-    {
-        return $this->responses->contains($request);
-    }
-
-    /**
-     * Checks if a request is failed
-     *
-     * @param RequestInterface $request
-     *
-     * @return boolean
-     */
-    public function isFailed(RequestInterface $request)
-    {
-        return $this->exceptions->contains($request);
     }
 
     /**
@@ -111,6 +81,8 @@ final class BatchResult
      * @param ResponseInterface $response
      *
      * @return BatchResult
+     *
+     * @internal
      */
     public function addResponse(RequestInterface $request, ResponseInterface $response)
     {
@@ -120,77 +92,8 @@ final class BatchResult
         return $new;
     }
 
-    /**
-     * Returns all exceptions
-     *
-     * @return Exception[]
-     */
-    public function getExceptions()
-    {
-        $exceptions = [];
-
-        foreach ($this->exceptions as $request) {
-            $exceptions[] = $this->exceptions[$request];
-        }
-
-        return $exceptions;
-    }
-
-    /**
-     * Returns an exception for a request or null if not found
-     *
-     * @param RequestInterface $request
-     *
-     * @return Exception|null
-     */
-    public function getExceptionFor(RequestInterface $request)
-    {
-        if ($this->exceptions->contains($request)) {
-            return $this->exceptions[$request];
-        }
-    }
-
-    /**
-     * Checks if there are any exceptions at all
-     *
-     * @return boolean
-     */
-    public function hasExceptions()
-    {
-        return $this->exceptions->count() > 0;
-    }
-
-    /**
-     * Checks if there is an exception for a request
-     *
-     * @param RequestInterface $request
-     *
-     * @return Exception
-     */
-    public function hasExceptionFor(RequestInterface $request)
-    {
-        return $this->exceptions->contains($request);
-    }
-
-    /**
-     * Adds an exception in an immutable way
-     *
-     * @param RequestInterface  $request
-     * @param Exception         $exception
-     *
-     * @return BatchResult
-     */
-    public function addException(RequestInterface $request, Exception $exception)
-    {
-        $new = clone $this;
-        $new->exceptions->attach($request, $exception);
-
-        return $new;
-    }
-
     public function __clone()
     {
         $this->responses = clone $this->responses;
-        $this->exceptions = clone $this->exceptions;
     }
 }
--- a/src/BatchResult.php
+++ b/src/BatchResult.php
@@ -2,6 +2,7 @@
 
 namespace Http\Client;
 
+use Http\Client\Exception\InvalidArgumentException;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
@@ -39,17 +40,21 @@ final class BatchResult
     }
 
     /**
-     * Returns a response of a request or null if not found
+     * Returns a response of a request
      *
      * @param RequestInterface $request
      *
-     * @return ResponseInterface|null
+     * @return ResponseInterface
+     *
+     * @throws InvalidArgumentException
      */
     public function getResponseFor(RequestInterface $request)
     {
-        if ($this->responses->contains($request)) {
-            return $this->responses[$request];
+        if (!$this->responses->contains($request)) {
+            throw new InvalidArgumentException('No response can be found for the given request');
         }
+
+        return $this->responses[$request];
     }
 
     /**
--- a/src/BatchResult.php
+++ b/src/BatchResult.php
@@ -1,0 +1,196 @@
+<?php
+
+namespace Http\Client;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Successful responses and exceptions returned from parallel request execution
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
+final class BatchResult
+{
+    /**
+     * @var \SplObjectStorage
+     */
+    private $responses;
+
+    /**
+     * @var \SplObjectStorage
+     */
+    private $exceptions;
+
+    public function __construct()
+    {
+        $this->responses = new \SplObjectStorage();
+        $this->exceptions = new \SplObjectStorage();
+    }
+
+    /**
+     * Checks if a request is successful
+     *
+     * @param RequestInterface $request
+     *
+     * @return boolean
+     */
+    public function isSuccessful(RequestInterface $request)
+    {
+        return $this->responses->contains($request);
+    }
+
+    /**
+     * Checks if a request is failed
+     *
+     * @param RequestInterface $request
+     *
+     * @return boolean
+     */
+    public function isFailed(RequestInterface $request)
+    {
+        return $this->exceptions->contains($request);
+    }
+
+    /**
+     * Returns all successful responses
+     *
+     * @return ResponseInterface[]
+     */
+    public function getResponses()
+    {
+        $responses = [];
+
+        foreach ($this->responses as $request) {
+            $responses[] = $this->responses[$request];
+        }
+
+        return $responses;
+    }
+
+    /**
+     * Returns a response of a request or null if not found
+     *
+     * @param RequestInterface $request
+     *
+     * @return ResponseInterface|null
+     */
+    public function getResponseFor(RequestInterface $request)
+    {
+        if ($this->responses->contains($request)) {
+            return $this->responses[$request];
+        }
+    }
+
+    /**
+     * Checks if there are any successful responses at all
+     *
+     * @return boolean
+     */
+    public function hasResponses()
+    {
+        return $this->responses->count() > 0;
+    }
+
+    /**
+     * Checks if there is a response of a request
+     *
+     * @param RequestInterface $request
+     *
+     * @return ResponseInterface
+     */
+    public function hasResponseFor(RequestInterface $request)
+    {
+        return $this->responses->contains($request);
+    }
+
+    /**
+     * Adds a response in an immutable way
+     *
+     * @param RequestInterface  $request
+     * @param ResponseInterface $response
+     *
+     * @return BatchResult
+     */
+    public function addResponse(RequestInterface $request, ResponseInterface $response)
+    {
+        $new = clone $this;
+        $new->responses->attach($request, $response);
+
+        return $new;
+    }
+
+    /**
+     * Returns all exceptions
+     *
+     * @return Exception[]
+     */
+    public function getExceptions()
+    {
+        $exceptions = [];
+
+        foreach ($this->exceptions as $request) {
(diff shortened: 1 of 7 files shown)


src/BatchResult.php
```diff
@@ -0,0 +1,196 @@
...
+     *
+     * @param RequestInterface $request
+     *
+     * @return ResponseInterface|null
+     */
+    public function getResponseFor(RequestInterface $request)
+    {
+        if ($this->responses->contains($request)) {
+            return $this->responses[$request];
+        }
```
dbu, line 57: otherwise, we should look in the exceptions to see if there is an exception carrying a response and return that. if that is not found, i would throw the ecxeption we find at $request. or an invalid argument exception if the request does not exist.

sagikazarmark, line 57: Currently responses are handled the following way: every response and exception is collected in the BatchResult. If there are any exceptions, The BatchException gets the result and it is thrown.

There is a getExceptionFor method to get the exception. If that exception is HttpException, you can get a response. I would not mix the two methods.

I also wouldn't throw the exception, because that way you cannot process the rest.

The reason why I put exceptions here as well is that an exception cannot be immutable. But it would possibly be better, to collect exceptions somewhere else (eg. right in the batch exception?)

src/Exception/BatchException.php
```diff
@@ -2,61 +2,49 @@
...
+     * Decides whether the result has any failures
      *
-     * @param TransferException $exception
+     * @param BatchResult $result
      *
-     * @return boolean TRUE if there is the exception else FALSE.
+     * @return BatchResult|BatchException
      */
-    public function hasException(TransferException $exception)
+    public static function decideReturnValue(BatchResult $result)
```
dbu: not sure if we need this. is the idea to avoid repeating logic in the adapters? while we are at it, we could have a second parameter to decide whether to throw the BatchException or return it. otherwise you would need an instanceof check if you want to throw, which would be weird.

and it would be even more weird to _return_ a BatchException from the client. 

oh, but apart from that, i think the decision of throw vs return can be moved to the adapter configuration. so we just say Client::sendRequests either returns a successful BatchResult and you can configure the client whether you want it to throw a BatchException or return a BatchResult. hm. which sounds like a huge pitfal and a pain for generic code that does not know whether the user will configure the client to throw or return... so with this i think we have to make up our mind and decide for either the exception or not.

my proposition is to say the client does the check for $result->hasExceptions() and if there are exceptions instantiates the BatchException and throws it. that way there is no ambiguity. and if i do have the same logic for both situations, in my application i do:

```
try {
    $batchResponse = $client->sendRequests(...);
} catch (BatchException $e) {
    $batchResponse = $e->getBatchResponse();
}
...
```

the alternative is that if i want an exception, the client would have to be written as

```
$batchResponse = $client->sendRequests(...);
if ($batchResponse->hasExceptions()) {
    throw new MyException;
}
...
```

both seem viable and not too much work. but i feel the throwing an exception is more consistent with the sendRequest signature.

src/BatchResult.php
sagikazarmark, line 57: I have an ugly idea for that. :stuck_out_tongue_winking_eye: 

In batch exception accept two objects: batch result, exception chain

The exception chain would be an object which accepts a request, an exception and the previous/next exception. THis way you have one object in the end to be injected into the batch exception. You can always get the next exception by calling getNextException, which returns the next in the chain.

dbu, line 57: hm. when i think of responses, i might want to process all responses,
even unsuccessful ones. but it indeed could get very confusing. so lets
adjust phpdoc to always say "successful responses" and throw an
exception if i try to request a response that does not exist (maybe with
a hint that the request failed, telling the user he should check with
the isSuccessful / isFailed method.

hannesvdvreken, line 57: > **throw** the exception we find at $request

I dunno about throwing it...

src/Exception/BatchException.php
sagikazarmark: > and it would be even more weird to return a BatchException from the client. 

You are right, I wasn't thinking when I wrote this.

I agree with the rest: if the result has exceptions, throw it in a batch exception.

It is also easier, because you have the same internal logic in sendRequest: ALWAYS collect responses and exceptions in BatchResult and it's not like: if we have an exception, we create a batchexception and put it there and in the end if we have a batchexception, we throw it.

src/BatchResult.php
dbu, line 57: yeah, throwing was a bad idea. i think separating is better. its probably safe to assume that most of the time the application needs different behaviour for failure and for success. or even does not need further processing of successes.

sagikazarmark, line 57: Damn, I wanted to write successful responses there. I knew you would notice it an I kept it in my mind.

sagikazarmark, line 57: One thing I am not really satisfied with though: we have all the exception methods in the result even if we know that we don't have any exceptions. So it feels like we would need two types of results: results with responses only and results with exceptions as well. But that's overcomplication.

src/Exception/BatchException.php
```diff
@@ -2,61 +2,49 @@
...
-            }
-        }
-
-        $this->exceptions = $exceptions;
+        $this->result = $result;
     }
 
     /**
-     * Returns all exceptions
+     * Returns successful responses or null
```
hannesvdvreken: Returns a BatchResult object which contains succesful responses and exceptions for failed requests.

src/BatchResult.php
dbu, line 57: hm. technically the BatchResponse could carry only the success
responses, and the failures could all be in BatchException. as we throw
the exception if there is any problem. then its clear: look for errors
in BatchException, look for successes in BatchResponse. and be aware
that the BatchResponse does not have all responses if you got it from
the BatchException. (which would be true for getResponse with the
current code as well, so its not more inconvenient, just more logic)

sagikazarmark, line 57: This is what I tried first. The problem is that an Exception cannot be cloned, so making it immutable that way doesn't work.

dbu, line 57: hm. i would prefer the clean and logic code structure over insisting on
the immutability. if people want to do weird stuff, they always find
ways to do it.

we can do an `@internal` annotation on the add method and note you
should not use this outside a client/adapter.

sagikazarmark, line 57: Should we do this in case of the BatchResult as well? I think it would be weird if the Result is immutable, the exception is not.

dbu, line 57: i would keep it immutable. its just a technical detail that the exception
is not forced immutable. it is to be considered immutable by the user

sagikazarmark, line 57: Done

src/BatchResult.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+namespace Http\Client;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Successful responses and exceptions returned from parallel request execution
```
dbu: s/and exceptions //

src/Exception/BatchException.php
```diff
@@ -2,61 +2,136 @@
 
 namespace Http\Client\Exception;
 
+use Http\Client\BatchResult;
+use Http\Client\Exception;
```
dbu, line 6: unused import

src/Exception/BatchException.php
```diff
@@ -2,61 +2,136 @@
...
+use Psr\Http\Message\RequestInterface;
+
 /**
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
-final class BatchException extends TransferException
+final class BatchException extends RuntimeException
```
dbu, line 16: lets document this: This exception is thrown when a batch of requests led to at least one failure. It holds the response/exception pairs and gives access to a BatchResult with the successful responses.

src/Exception/BatchException.php
```diff
@@ -2,61 +2,136 @@
...
+    /**
+     * Checks if a request is successful
+     *
+     * @param RequestInterface $request
+     *
+     * @return boolean
+     */
+    public function isSuccessful(RequestInterface $request)
+    {
+        return isset($this->result) && $this->result->hasResponseFor($request);
```
dbu: i would say the client MUST set the BatchResult so that we can skip the isset check.

src/Exception/BatchException.php
sagikazarmark, line 6: Which is unused?

src/Exception/BatchException.php
```diff
@@ -2,61 +2,136 @@
...
-     * @return boolean TRUE if there is the exception else FALSE.
+     * @return Exception|null
      */
-    public function hasException(TransferException $exception)
+    public function getExceptionFor(RequestInterface $request)
     {
-        return array_search($exception, $this->exceptions, true) !== false;
+        if ($this->exceptions->contains($request)) {
+            return $this->exceptions[$request];
+        }
```
dbu, line 122: lets throw an InvalidArgumentException if you request the exception for a request that is not found.

src/Exception/BatchException.php
```diff
@@ -2,61 +2,136 @@
...
     /**
-     * Checks if any exception exists
+     * Checks if there are any exceptions at all
      *
      * @return boolean
      */
     public function hasExceptions()
     {
-        return !empty($this->exceptions);
+        return $this->exceptions->count() > 0;
```
dbu: i think we can drop this. the class is not created if there are no exceptions.

src/Exception/BatchException.php
```diff
@@ -2,61 +2,136 @@
...
     private $exceptions;
 
+    public function __construct()
+    {
+        $this->exceptions = new \SplObjectStorage();
+    }
+
     /**
-     * @param TransferException[] $exceptions
+     * Returns a list BatchResult which contains succesful responses
```
dbu: Returns the BatchResult that contains those responses that where successful.

Note that the BatchResult may contains 0 responses if all requests failed.

src/Exception/BatchException.php
sagikazarmark: It is just a foolproofness check. The best solution would be to inject the BatchResult in the constructor, but it is not possible if the order or processing is linear, since the BatchResult is immutable, so after injecting we cannot modify the internal object. One workaround for this could be to create and empty BatchResult in the exception constructor.

dbu: or possibly use $this->getResult() and make getResult() create an empty BatchResult if none is set.

src/Exception/BatchException.php
sagikazarmark: Agree.

src/Exception/BatchException.php
```diff
@@ -2,61 +2,136 @@
...
+    /**
+     * Sets the successful response list
+     *
+     * @param BatchResult $result
+     *
+     * @internal
+     */
+    public function setResult(BatchResult $result)
+    {
+        $this->result = $result;
```
dbu, line 64: to make this better immutable, we could add (before this line)

```
if ($this->result) {
     throw new \InvalidArgumentException('BatchResult is already set');
}
```

src/Exception/BatchException.php
sagikazarmark: :+1:

src/HttpPsrClient.php
```diff
@@ -30,9 +31,10 @@ public function sendRequest(RequestInterface $request, array $options = []);
      * @param RequestInterface[] $requests
      * @param array              $options
      *
-     * @return ResponseInterface[]
+     * @return BatchResult
      *
      * @throws Exception
+     * @throws BatchException
```
dbu, line 40: add a description: "If one or more requests led to an exception, the BatchException is thrown. The BatchException also gives access to the BatchResult for the successful responses."

src/HttpPsrClient.php
```diff
@@ -30,9 +31,10 @@ public function sendRequest(RequestInterface $request, array $options = []);
      * @param RequestInterface[] $requests
      * @param array              $options
      *
-     * @return ResponseInterface[]
+     * @return BatchResult
```
dbu: description "If all requests where successful."

src/Exception/BatchException.php
dbu, line 6: i thought the exception interface. but actually its used. sorry.

src/Exception/BatchException.php
sagikazarmark, line 64: It is possible. But every `getResult` when no result is set will have to create a new object. It would be easier if a new object is created once and then it is set as the actual result.

src/HttpPsrClient.php
sagikazarmark, line 40: Isn't it too long? The BatchException API contains this information (about the result).

src/Exception/BatchException.php
dbu, line 64: no, getResult can be implemented as

```
if (!$this->result) {
    $this->result = new BatchResult(); // or also $this->setResult(new BatchResult()); as the result is not yet set
}

return $this->result;
```

src/HttpPsrClient.php
dbu, line 40: i think it would be useful. but we can also skip the second sentence, or put it into the general method description.

src/Exception/BatchException.php
sagikazarmark, line 64: This is how I implemented it. But in this case, the code in the setter you wrote won't work.

dbu, line 64: why not? the getter only creates an object if the result is not set. the setter only complains if the result is already set.

sagikazarmark, line 64: You are right. I forgot that `getResult` will probably not be called during response processing.

src/Exception/BatchException.php
sagikazarmark, line 122: I am not sure about this. I always tend to rather return null than throwing an exception. Is it more elegant to throw one?

sagikazarmark, line 122: Also, we should be consistent and do the same in batch result.

dbu, line 122: returning null is a horrible thing imho :-)
as it makes errors go unnoticed. i rather have an exception with a
stacktrace than "fatal error: called method on something that is not an
object"

dbu, line 122: yes, exceptions whenever something weird is being done imho

sagikazarmark, line 122: SplObjectStorage throws UnexpectedValueException in case an object is not found. Probably we could catch that exception and throw ours.

joelwurtz:
Maybe providing a Trait for the method sendRequests on the HttpPsrClient (using the sendRequest method) would be useful for client/adapter not handling this use case ?

``` php
trait BatchRequest {
    public function sendRequests($requests)
    {
         foreach ($requests as $request) {
             ...
         }
    }
}
```

sagikazarmark:
@joelwurtz Might be useful in cases where the underlying client does not support parallel requests. Internally it could rely on `sendRequest` method of the interface. Would you like to open a PR against the terminology branch?

dbu:
the idea makes sense. i just wonder how much implementation this
repository should provide. maybe we could have an adapter-base (or
something like that) repository that provides a trait and other utility
things we come up with. the trait not on the same level as the spec itself.

sagikazarmark:
Yeah, that's a good point as well.

I would say, until we don't have enough utilities, collect them at one place. Before release, we can separate components which are not part of the spec. The utility namespace could even be the same or we can add the Util word to the namespace. WDYT?

joelwurtz:
No problem for making a PR @sagikazarmark will wait to have this merged before doing it (so no changes need to be made with the new BatchResult object)

sagikazarmark:
Actually, I think this is ready to merge. @dbu ?

dbu:
agreed. and lets have a Util namespace or something right away

sagikazarmark:
@joelwurtz PR merged. If you provide a PR, please do that under the Util namespace.

PR #60, opened 2015-09-29
Remove BatchResult implementation, remove custom SPL exceptions

sagikazarmark:
I removed the BatchResult implementation from the main repo, it is going to be in the utility repo.

Some other major changes in this PR:

I removed custom SPL exception extensions.

Reasoning:

The reasons why we have an `Exception` interface are:
- have a common type for our domain exceptions, since exceptions must extend an Exception class
- allow to catch any exceptions thrown by the adapter

However doing so only make sense when one of the domain exceptions is thrown. For example BatchException. It can be caught by catching the Exception interface. However, exceptions thrown by that BatchException (RuntimeException, UnexpectedValueException) should be handled locally, because they are usually because some sort of logic issue (like getResult called before it is set).

So in this PR I removed all custom SPL exceptions and started to use the core ones. If we ever need more exceptions, like we have now, we can always create new domain exceptions.

What do you think guys?


sagikazarmark:
There is one question though:

Currently the BatchException only accepts instances of our Exception interface. Should it accept any kind of exceptions? Of course exceptions should be converted to ours if possible, and I can't really imagine any cases where it wouldn't. Still, we should consider this question. (But I rather think we should keep this restriction)

hannesvdvreken:
:+1:

src/Exception/InvalidArgumentException.php
```diff
@@ -1,14 +0,0 @@
...
-namespace Http\Client\Exception;
-
-use Http\Client\Exception;
-
-/**
- * Custom Invalid Argument exception
- *
- * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
- */
-final class InvalidArgumentException extends \InvalidArgumentException implements Exception
```
dbu, line 12: removing this means a user can not catch this specific exception anymore. but as this is a programming error we might not need it.

src/Exception/UnexpectedValueException.php
```diff
@@ -1,14 +0,0 @@
...
-
-use Http\Client\Exception;
-
-/**
- * Custom Unexpected Value exception
- *
- * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
- */
-final class UnexpectedValueException extends \UnexpectedValueException implements Exception
-{
```
dbu, line 13: removing this means a user can not catch this specific exception anymore. are there valid use cases to want catch this one?

src/Exception/InvalidArgumentException.php
sagikazarmark, line 12: That's exactly what I tried to point in the PR description.

src/Exception/UnexpectedValueException.php
sagikazarmark, line 13: This was only used in the BatchException and BatchResult. This is thrown in case of a "programming error" as you've said.

src/Exception/InvalidArgumentException.php
dbu, line 12: sorry, did not read well enough. yeah lets do that. can you link the place where you add the BatchResult in the utilities repo?

sagikazarmark, line 12: As said earlier I will only do that once the utility separation is ready to merge, so that it is clear what needs to go into the util repo.

PR #61, opened 2015-09-29
Move exceptions to BatchResult

sagikazarmark:
I am coming up with this solution again. If we decide to merge #60, we should consider merging this one as well.

Based on how the [BatchRequest trait](https://github.com/php-http/adapter/blob/master/src/Util/BatchRequest.php) is created, it turns out that the current BatchResult-BatchException solution is not really efficient.

Moving everything to the BatchResponse (again) makes it easier to collect everything in one place and think later about exceptions.

This solution also allows to wrap existing batch objects in cases a client supports parallel execution (like in case of guzzle6)

I also think that the API is easier to read this way. We can also forget about the `@internal` docblock we added because Exceptions cannot be cloned.

To sum up, there are plenty of advantages to move exceptions to the BatchResult object.

What do you think?

### DIFF
--- a/spec/Exception/BatchExceptionSpec.php
+++ b/spec/Exception/BatchExceptionSpec.php
@@ -9,6 +9,11 @@ use PhpSpec\ObjectBehavior;
 
 class BatchExceptionSpec extends ObjectBehavior
 {
+    function let(BatchResult $batchResult)
+    {
+        $this->beConstructedWith($batchResult);
+    }
+
     function it_is_initializable()
     {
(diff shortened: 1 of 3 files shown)


hannesvdvreken:
:+1:

dbu:
yeah, really looks simpler. so lets do this. should we keep a BatchException that can be used when at least one request in the batch failed and that gives access to the BatchResult? or do we just say when you batch then you need to manually check each response as there never ever will be exceptions thrown?

sagikazarmark:
I would say the BatchException should still be there.

PR #62, opened 2015-09-29
document options argument

### LINKED ISSUE #49: List of possible options
We need to have a list of possible values for the `options` parameter. These are general options which should work with ALL the clients.

Possible options:
- timeout [0]
- exceptions [false]: whether the user wants 4xx and 5xx responses to be converted to exceptions
- base_uri [null]: from #36

dbu:
fix #49

### DIFF
--- a/src/HttpMethodsClient.php
+++ b/src/HttpMethodsClient.php
@@ -11,6 +11,14 @@ use Psr\Http\Message\ResponseInterface;
  *
  * Use this interface when you do not have PSR-7 RequestInterface instances available.
  *
+ * Implementations should be configured in their constructor. For runtime behaviour changes, there
+ * is an $options parameter. It SHOULD be validated by the client and an exception thrown if
+ * unknown options are passed.
+ * Option that must be accepted by every client implementation (though it is free to ignore them
+ * if the underlying implementation does not support the option):
+ *
+ *   - timeout (int): the timeout for the HTTP request in seconds. 0 means to use the default.
+ *
  * @author Márk Sági-Kazár mark.sagikazar@gmail.com>
  */
 interface HttpMethodsClient
--- a/src/HttpPsrClient.php
+++ b/src/HttpPsrClient.php
@@ -9,6 +9,14 @@ use Psr\Http\Message\ResponseInterface;
 /**
  * Sends one or more PSR-7 Request
  *
+ * Implementations should be configured in their constructor. For runtime behaviour changes, there
+ * is an $options parameter. It SHOULD be validated by the client and an exception thrown if
+ * unknown options are passed.
+ * Option that must be accepted by every client implementation (though it is free to ignore them
+ * if the underlying implementation does not support the option):
+ *
+ *   - timeout (int): the timeout for the HTTP request in seconds. 0 means to use the default.
+ *
  * @author GeLo <geloen.eric@gmail.com>
  */
 interface HttpPsrClient


sagikazarmark:
:+1:

src/HttpMethodsClient.php
```diff
@@ -11,6 +11,14 @@
  *
  * Use this interface when you do not have PSR-7 RequestInterface instances available.
  *
+ * Implementations should be configured in their constructor. For runtime behaviour changes, there
+ * is an $options parameter. It SHOULD be validated by the client and an exception thrown if
+ * unknown options are passed.
```
joelwurtz, line 16: > exception thrown if unknown options are passed.

How to handle specific options for an adapter then ?

dbu, line 16: > ```
> exception thrown if unknown options are passed.
> ```
> 
> How to handle specific options for an adapter then ?

only pass them when you know you are using that specific client/adapter.
in a library, such options - if needed at all - should be injected.

to be honest i am a bit unsure about this. it would be cleaner to not
have options at all. but it could lead to an excessive number of
adapter/client instances to have them with all the various options.
maybe we should still do it that way.

dbu, line 16: removing options parameter would force people do to clean programming
and inject a properly configured adapter/client into their code.

sagikazarmark, line 16: When we started this project we had a ConfigurableRequest, which wrapped the original request and had a `getOptions` method.

To be honest: how often do you need different timeout option in a client?

joelwurtz, line 16: Maybe not different timeout, but other behaviour may be needed, like sometimes you want an Async Request/Response and sometimes not.

sagikazarmark, line 16: Well, the SocketClient might be special from this point of view. These specific settings should probably be set in the constructor, or using specific setters.

joelwurtz, line 16: I don't agree, we also need this settings to be set in the request call,

As an exemple, for our library (docker-php) we need to have this settings per request, as i want to have a logs response in a async manner (it's a chunked reponse which is populated from consecutive actions and stay alive "forever"), but we also want to have the stop / start calls in a sync manner (to be sure that it is correctly stop / start before doing anything else).

But i agree with @dbu and we should only inject these options when we know we use a specific client.

Another approach would be to have a "namespaced" option name for each adapter, where all the configuration under this namespace should only be handle by the adapter

``` php
// Global / Interoperable option
$options['timeout'] = 30;
// Specific option for socket adapter
$options['socket-adapter'] = [
    'specific-otption-for-socket' => 'value'
];
// Specific option for guzzle6 adapter
$options['guzzle6-adapter'] = [
    'specific-otption-for-guzzle6' => 'value'
];
```

sagikazarmark, line 16: I would rather leave this to the code using the client and not add things like namespaces.

sagikazarmark:
@dbu Should we have an UnkownOptionException in the contract?

src/HttpMethodsClient.php
dbu, line 16: > As an exemple, for our library (docker-php) we need to have this
> settings per request, as i want to have a logs response in a async
> manner (it's a chunked reponse which is populated from consecutive
> actions and stay alive "forever"), but we also want to have the stop /
> start calls in a sync manner (to be sure that it is correctly stop /
> start before doing anything else).

i think this is exactly an example where you could just use two client
instances: a synchronous one for the start/stop and an asynchronous for
the logging. that way is a lot more inversion of control. your library
specifies that it needs a sync client and one that is efficient for
logging. it does not need to know the options to make that happen. the
application bootstrap just creates the two clients and gives them to
your library.

i think i am for entirely dropping the $options argument. it is just an
invitation to write client specific code, and the purpose of this
abstraction is to avoid that.

PR #64, opened 2015-10-01
Drop options from interfaces

### LINKED ISSUE #63: Drop options argument
In a previous PR we started to think with @dbu that we should drop the options argument. So far we could only found the `timeout` option which could be passed there. We would like to prevent that options is used for client specific things.

Actually timeout is something that might not be needed to be set for every request, but configured in the constructor.

So the current proposal is to drop the options parameter entirely.

sagikazarmark:
Fixes #63


dbu:
:+1:

PR #67, opened 2015-10-01
mention ivory in the readme

dbu:
to avoid confusion. ideally, @egeloen would add some note in ivory about this as well. (for now maybe with a warning that this one is not yet ready for production)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ $ composer require php-http/httplug
 
 This is the contract package for HTTP Client. Use it to create HTTP Clients which are interoperable and compatible with [PSR-7](http://www.php-fig.org/psr/psr-7/).
 
+This library is the official successor of the [ivory http adapter](https://github.com/egeloen/ivory-http-adapter).
 
 ## Documentation
 


sagikazarmark:
Be careful with the name 'Ivory'. It is a vendor name as well, not just the name for the HTTP Adapter package.

dbu:
better?

willdurand:
:+1:

egeloen:
I just added a note in the readme about the deprecation and these set of packages.

egeloen:
I think I will deprecate it officially in composer when the 1.0 is out here.

README.md
```diff
@@ -25,6 +25,7 @@ $ composer require php-http/httplug
 
 This is the contract package for HTTP Client. Use it to create HTTP Clients which are interoperable and compatible with [PSR-7](http://www.php-fig.org/psr/psr-7/).
 
+This library is the official successor of the [ivory http adapter](https://github.com/egeloen/ivory-http-adapter) and the code is based on the ivory http adapter code.
```
sagikazarmark: Ivory http adapter and code expressions are repeated here. Can we make it simpler? Also, the code is completely refactored. Actually it has been refactored several times. I think it is enough to mention that the idea is from ivory, it evolved from an adapter package to a PSR-7 HTTP Client abstraction and it is a successor of IHA.

dbu:
thanks a lot egeloen!

i updated the text a bit. ok @sagikazarmark ?

sagikazarmark:
:+1:

PR #69, opened 2015-10-20
move HttpMethodsClient to implementation in utils, rename psr client

dbu:
Continue the utility separation effort. Moving to utility in https://github.com/php-http/utility/pull/2


sagikazarmark:
:+1:

PR #73, opened 2015-10-23
cleanup phpdoc and BatchResult interface

dbu:
mainly adjusting the exceptions, but i also went over the BatchResult interface. we had duplicated methods and i reordered them to be consistent.

isSuccessful and hasResponseFor are the same
as are isFailed and hasExceptionFor

i went with isSuccessful and isFailed, but maybe hasExceptionFor / hasResponseFor would be more straightforward. wdyt?

### DIFF
--- a/spec/Exception/TransferExceptionSpec.php
+++ b/spec/Exception/TransferExceptionSpec.php
@@ -2,14 +2,14 @@
 
 namespace spec\Http\Client\Exception;
 
+use Http\Client\Exception\TransferException;
 use PhpSpec\ObjectBehavior;
-use Prophecy\Argument;
 
 class TransferExceptionSpec extends ObjectBehavior
 {
-    function it_is_initializable()
+    function let()
     {
-        $this->shouldHaveType('Http\Client\Exception\TransferException');
+        $this->beAnInstanceOf('spec\Http\Client\Exception\TransferExceptionStub');
     }
 
     function it_is_a_runtime_exception()
@@ -21,4 +21,8 @@ class TransferExceptionSpec extends ObjectBehavior
     {
         $this->shouldImplement('Http\Client\Exception');
     }
+}
+
+class TransferExceptionStub extends TransferException
+{
 }
--- a/src/Exception/RequestException.php
+++ b/src/Exception/RequestException.php
@@ -5,7 +5,10 @@ namespace Http\Client\Exception;
 use Psr\Http\Message\RequestInterface;
 
 /**
- * Base exception for when a request failed.
+ * Exception for when a request failed, providing access to the failed request.
+ *
+ * This could be due to an invalid request, or one of the extending exceptions
+ * for network errors or HTTP error responses.
  *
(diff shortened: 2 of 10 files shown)


src/Exception/TransferException.php
```diff
@@ -9,6 +9,6 @@
  *
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
-class TransferException extends \RuntimeException implements Exception
+abstract class TransferException extends \RuntimeException implements Exception
```
dbu, line 12: i think we never need to directly instantiate this exception. or do we?

src/Exception/RequestException.php
```diff
@@ -37,19 +40,4 @@ public function getRequest()
...
         return $this->request;
     }
-
-    /**
-     * @param RequestInterface $request
-     * @param \Exception       $e
-     *
-     * @return RequestException
-     */
-    public static function wrapException(RequestInterface $request, \Exception $e)
```
dbu, line 47: not sure what this was for. i guess we should go for the correct exception and just use new. its also weird how the passed request would get lost if $e is already a RequestException...

src/Exception/TransferException.php
sagikazarmark, line 12: Don't think so.

sagikazarmark:
I think isSuccessful and isFailed are better from a domain point of view. We are curious about if the request was successful or failed after all.

dbu:
BatchException never neds to have a previous exception, right? it currently does not have that option, but it also would not make much sense.

src/Exception/RequestException.php
sagikazarmark, line 47: It's from guzzle, free to remove it.

sagikazarmark:
Agreed. Actually there are several previous exceptions which are exposed through a custom API.

spec/Exception/TransferExceptionSpec.php
```diff
@@ -1,24 +0,0 @@
-<?php
-
-namespace spec\Http\Client\Exception;
-
-use PhpSpec\ObjectBehavior;
-use Prophecy\Argument;
-
-class TransferExceptionSpec extends ObjectBehavior
```
dbu, line 8: or is there a way to spec these things without instantiating the class?

dbu:
this looks ready from my point, please merge if you agree. i added https://github.com/php-http/utils/pull/12 to adjust the BatchResult implementation.

spec/Exception/TransferExceptionSpec.php
sagikazarmark, line 8: You can add a stub, like in case of BatchRequest trait and test that.

sagikazarmark:
Do you want to add tests for TransferException prior to merging?

dbu:
added the TransferExcpetion spec back. ready to merge i hope?

sagikazarmark:
Go ahead once tests are finished.

PR #77, opened 2015-10-26
Async Http Client

joelwurtz:
See #74 and #75 discussion for a more detailed description.

This PR intend to add asynchronous support for httplug by adding a new interface `HttpAsyncClient` which allow to send an asynchronous request and return a promise for the response.

The current Promise interface follow specification of https://promisesaplus.com/ from an user point of view (there is no method for fulfilling or rejecting the promise, it should be up to the implementation user only care on how to handle the return and not on how to resolve the promise)

### DIFF
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -44,7 +44,7 @@ interface Promise
      *
      * @param callable $onFulfilled Called when a response will be available.
      *
-     * It will receive a Psr\Http\Message\RequestInterface object as the first argument
+     * It will receive a Psr\Http\Message\ResponseInterface object as the first argument
      * If the callback is null it must not be called.
      * It must be called after promise is fulfilled, with promise’s value (ResponseInterface) as its first argument.
      * It must not be called before promise is fulfilled.
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -8,6 +8,10 @@ use Psr\Http\Message\ResponseInterface;
  * Promise represents a response that may not be available yet, but will be resolved at some point in future.
  * It acts like a proxy to the actual response.
  *
+ * This interface is an extension of the promises/a+ specification https://promisesaplus.com/
+ * Value is replaced by an object where its class implement a Psr\Http\Message\RequestInterface.
+ * Reason is replaced by an object where its class implement a Http\Client\Exception.
+ *
  * @author Joel Wurtz <joel.wurtz@gmail.com>
  */
 interface Promise
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -18,5 +18,5 @@ interface Promise
      *
      * @return Promise Always returns a new promise which is resolved with value of the executed callback (onFulfilled / onRejected)
      */
-    public function then(callable $onFulfilled = null, callable $onRejected = null);
+    public function then(callable $onFulfilled, callable $onRejected);
 }
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -10,13 +10,31 @@ namespace Http\Client;
  */
 interface Promise
 {
+    const PENDING   = 0;
+    const FULFILLED = 1;
+    const REJECTED  = 2;
+
     /**
      * Add behavior for when the promise is resolved or rejected (response will be available, or error happens)
      *
-     * @param callable $onFulfilled Called when a response will be available, it will receive a Psr\Http\Message\RequestInterface object as the first argument
-     * @param callable $onRejected  Called when an error happens, it will receive a Http\Client\Exception object as the first argument
+     * @param callable $onFulfilled Called when a response will be available.
+     *
+     * It will receive a Psr\Http\Message\RequestInterface object as the first argument
+     * If the callback is null it should not be called.
+     *
+     * @param callable $onRejected  Called when an error happens.
+     *
+     * It will receive a Http\Client\Exception object as the first argument.
+     * If the callback is null it should not be called.
      *
      * @return Promise Always returns a new promise which is resolved with value of the executed callback (onFulfilled / onRejected)
      */
-    public function then(callable $onFulfilled, callable $onRejected);
+    public function then(callable $onFulfilled = null, callable $onRejected = null);
+
+    /**
+     * Get the state of the promise, one of PENDING, FULFILLED or REJECTED
+     *
+     * @return int
+     */
+    public function getState();
 }
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -38,7 +38,7 @@ interface Promise
      * The callback will be called when the response or exception arrived and never more than once.
      *
      * @param callable $onFulfilled Called when a response will be available.
-     * @param callable $onRejected Called when an error happens.
+     * @param callable $onRejected  Called when an error happens.
      *
      * You must always return the Response in the interface or throw an Exception.
      *
@@ -56,27 +56,28 @@ interface Promise
     /**
      * Return the value of the promise (fulfilled).
      *
-     * @throws \LogicException When the promise is not fulfilled.
-     *
      * @return ResponseInterface Response Object only when the Promise is fulfilled.
+     *
+     * @throws \LogicException When the promise is not fulfilled.
      */
     public function getResponse();
 
     /**
      * Return the reason of the promise (rejected).
      *
-     * @throws \LogicException When the promise is not rejected.
-     *
      * @return Exception Exception Object only when the Promise is rejected.
+     *
+     * If the exception is an instance of Http\Client\Exception\HttpException it will contain
+     * the response object with the status code and the http reason.
+     *
+     * @throws \LogicException When the promise is not rejected.
      */
-    public function getError();
+    public function getException();
 
     /**
      * Wait for the promise to be fulfilled or rejected.
      *
-     * This function does not return a result, it simply wait for response or error
-     * of the request to be available, change the state of the promise and call one
-     * of the then callable.
+     * When this method returns, the request has been resolved and the appropriate callable has terminated.
      */
     public function wait();
 }
--- a/src/Promise.php
+++ b/src/Promise.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Http\Client;
+
+/**
+ * Promise represents a response that may not be available yet, but will be resolved at some point in future.
+ * It acts like a proxy to the actual response.
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+interface Promise
+{
+    /**
+     * Add behavior for when the promise is resolved or rejected (response will be available, or error happens)
+     *
+     * @param callable $onFulfilled Called when a response will be available, it will receive a Psr\Http\Message\RequestInterface object as the first argument
+     * @param callable $onRejected  Called when an error happens, it will receive a Http\Client\Exception object as the first argument
+     *
+     * @return Promise Always returns a new promise which is resolved with value of the executed callback (onFulfilled / onRejected)
+     */
+    public function then(callable $onFulfilled = null, callable $onRejected = null);
+}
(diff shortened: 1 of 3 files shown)


joelwurtz:
Actually the onFulfilled / onRejected are a callable, WDYT if we replace them by an interface ? 

(Would be easier to typehint and also not so a problem with inline implementation for the futur, as we will have anonymous class, and we can always wrap a defaut implementation which takes a callable)

sagikazarmark:
Default implementation on the other hand causes to depend on the utils package.

joelwurtz:
Hum right, we can stick with callable for the moment.

I think it is ready for review then, i don't see anything to add. I will maybe add some docs on the getState method as in my mind this method as the role to wait / check the underlying implementation (like a stream_select).

src/Promise.php
```diff
@@ -0,0 +1,85 @@
...
+     * It must not be called before promise is fulfilled.
+     * It must not be called more than once.
+     *
+     * @param callable $onRejected Called when an error happens.
+     *
+     * It will receive a Http\Client\Exception object as the first argument.
+     * If the callback is null it must not be called.
+     * It must be called after promise is rejected, with promise’s reason (Exception) as its first argument.
+     * It must not be called before promise is rejected.
+     * It must not be called more than once.
```
dbu: this looks like phpspec to me. maybe we can condense this a bit and make it more readable? and i would write it from the perspective of somebody using the promise, not the implementors.

the general description could state what is common:
- If you do not care about one of the cases, you can set the corresponding callable to null
- The callback will be called when the response resp. exception arrived and never more than once.

...

src/Promise.php
```diff
@@ -0,0 +1,85 @@
...
+     * Get the state of the promise, one of PENDING, FULFILLED or REJECTED
+     *
+     * @return int
+     */
+    public function getState();
+
+    /**
+     * Return the value of the promise (fulfilled)
+     *
+     * @return null|ResponseInterface Null if the promise is rejected or pending, the response object otherwise
```
dbu: i would suggest to say we throw an exception if the promise was rejected.

src/Promise.php
sagikazarmark: Agree

src/Promise.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public function getResponse();
+
+    /**
+     * Return the reason of the promise (rejected)
+     *
+     * @return null|Exception Null if the promise is fulfilled or pending, the exception object otherwise
```
dbu: i would suggest to say we throw an exception if the promise was fulfilled. maybe even when its pending. there is getState to check the state.

src/Promise.php
dbu: and also if its pending. there is getState for checking the state

dbu:
very cool! 

in terms of usability, i wonder if this client could extend the normal HttpClient. when you can send async requests, you for sure can also send other requests (worst case waiting for the result of an async request in the client). or not? just trying to think how many if instanceof we want users of the library to do...

joelwurtz:
Yes i was thinking the same thing, itreally depend on implementation as guzzle provide both mecanism we don't need to do that, but for React adapter i think we will provide also a "wait" mechanism for the synchronous sendRequest

src/Promise.php
joelwurtz: Agreed also, as i want to limit all these to the user point of view, this should be more in the implementation documentation

src/Promise.php
joelwurtz: Which exception should we throw ? \UnexpectedValueException ?

dbu: its not about the value, its about the state. maybe LogicException or
BadMethodCallException

sagikazarmark:
I think interface inheritance does not really matter here.

dbu:
yeah, actually a use case where you need both async _and_ sync request seems not very obvious. guess we are fine like this, keeps things simpler.

so lets clean up the exception behaviour and callback documentation and then merge this. oh, and maybe build a doc PR that goes with this ;-)

sagikazarmark:
:+1:

joelwurtz:
Just updated but dot not merge it, i want to try react and guzzle6 implementations before submitting this, to be sure that this interface has all requirements. Also i will do the documentation PR linked to this.

sagikazarmark:
Won't merge it until the WIP tag is included in the title.

joelwurtz:
After looking at the Guzzle and React implementation for their async request i think we need a wait function to this interface.

As in real, expect with some "hacking" (like using the tick register handler of PHP), there is no way that our callback will be called magically if somehow we do not wait for the result to be available at some moment.

So IMO we must provide a wait function for the user when he needs the response of this request to be complete.

dbu:
agreed. can we just use the guzzle impl in the guzzle adapter?

joelwurtz:
We have to provide a decorator to not collide with their interface, but this is pretty straigthforward otherwise.

dbu:
> We have to provide a decorator to not collide with their interface, but
> this is pretty straigthforward otherwise.

:+1: decorator is more clean than extend to add the interface anyways.

src/Promise.php
```diff
@@ -0,0 +1,82 @@
...
+    const REJECTED  = "rejected";
+
+    /**
+     * Add behavior for when the promise is resolved or rejected (response will be available, or error happens).
+     *
+     * If you do not care about one of the cases, you can set the corresponding callable to null
+     * The callback will be called when the response or exception arrived and never more than once.
+     *
+     * @param callable $onFulfilled Called when a response will be available.
+     * @param callable $onRejected Called when an error happens.
```
sagikazarmark: Description indentation

src/Promise.php
```diff
@@ -0,0 +1,82 @@
...
+     * @throws \LogicException When the promise is not fulfilled.
+     *
+     * @return ResponseInterface Response Object only when the Promise is fulfilled.
+     */
+    public function getResponse();
+
+    /**
+     * Return the reason of the promise (rejected).
+     *
+     * @throws \LogicException When the promise is not rejected.
```
sagikazarmark: Order of phpdoc: return, throws

sagikazarmark:
Not sure if I completely get this async idea. Shouldn't calling the wait return the response or throw an exception?

joelwurtz:
Guzzle does that by default but also provide a way to do like here (using the $unwrap parameter)

In our case, i prefer to have a limited ways of getting the response, (this also make implementation lighter).

But i really don't mind about one solution or the other.

src/Promise.php
```diff
@@ -0,0 +1,82 @@
...
+     * @return Exception Exception Object only when the Promise is rejected.
+     */
+    public function getError();
+
+    /**
+     * Wait for the promise to be fulfilled or rejected.
+     *
+     * This function does not return a result, it simply wait for response or error
+     * of the request to be available, change the state of the promise and call one
+     * of the then callable.
```
dbu: lets add something like "When this method returns, the request has been resolved and the appropriate callable has terminated.

dbu:
hm, i agree we only want one or the other way. not sure which is better though. when using the callable mechanism, not returning anything makes sense. when not using them, returning would make more sense. 

but i guess the callable mechanism is really useful, particularly when there are several promises and we don't know which will be resolved first. (does that work with php? when i wait on one promise, if another one resolves while i wait, will the callable of the other promise be executed?)

if callables work that way, i would prefer that we promote the callable approach in the doc and have wait() not return anything. we should keep getError / getResponse in either case, to make it not unnecessary hard to access that information if desired (like write a callable to deposit the information somewhere...)

src/Promise.php
```diff
@@ -0,0 +1,82 @@
...
+    public function getResponse();
+
+    /**
+     * Return the reason of the promise (rejected).
+     *
+     * @throws \LogicException When the promise is not rejected.
+     *
+     * @return Exception Exception Object only when the Promise is rejected.
+     */
+    public function getError();
```
dbu: why not `getException`?

src/Promise.php
```diff
@@ -0,0 +1,82 @@
...
+     * Return the value of the promise (fulfilled).
+     *
+     * @throws \LogicException When the promise is not fulfilled.
+     *
+     * @return ResponseInterface Response Object only when the Promise is fulfilled.
+     */
+    public function getResponse();
+
+    /**
+     * Return the reason of the promise (rejected).
```
dbu, line 66: "Return the reason why the promise was rejected."?

and we could mention that if this exception is a HttpException, it will contain the Response object with the error status code inside. (in the hope that people will be less confused when trying to figure out how to get to that information)

joelwurtz:
> but i guess the callable mechanism is really useful, particularly when there are several promises and we don't know which will be resolved first. (does that work with php? when i wait on one promise, if another one resolves while i wait, will the callable of the other promise be executed?)

What do you mean by several promises ?

A promise for another sendAsyncRequest ?
Or a chained one like  : 

```
$promise
  ->then(function() ..., function() ...)
  ->then(function () ..., function () ...)
;
```

If it's a promise for another sendAsyncRequest, it will depend on the implementation, for guzzle i don't think so but for react certainly if we use a global event loop.

For the chained one it will call promises in the order of declaration.

By the way even if we return a Response / Exception in the wait method, the `then` callables will always be called as it is mandatory.

dbu:
i was thinking about parallel async requests, yes. so lets not return anything in the `wait` and promote the callables way, but keep the getResponse / getException methods for people that want to work differently.

joelwurtz:
So, this should be good to merge, after reviewing the Observable vs Promise discussion on the Promise RFC on fig group, i don't think this pattern is better or not (i find it very similar, see discussion on slack with @sagikazarmark).

src/Promise.php
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @return Exception Exception Object only when the Promise is rejected.
+     *
+     * If the exception is an instance of Http\Client\Exception\HttpException it will contain
+     * the response object with the status code and the http reason.
```
dbu, line 71: i think this should go above the `@return` line

src/Promise.php
dbu, line 66: what do you think about the wording "Get the reason why the promise was rejected." ?

dbu:
how soon can we expect FIG to release a promise standard? if its reasonable to expect it by end of the year, i would delay support for async to 1.1 of HTTPlug to get their promises rather than define our own.

alternatively i could live with clearly stating that the Promise interface is to be considered experimental and will be replaced by the FIG one in a minor version change. or we just go 2.0 after 3 months... its hard to say i guess how long FIG will take. and we really want to wait with adopting their thing until its actually stable released. maybe a version 2.0 of HTTPlug in half a year is the way to go.

sagikazarmark:
I don't think we can expect a standard this year.

What if we place the async stuff in a separate repository as @joelwurtz suggested. We can have it experimental until the Promise question is cleared, we can merge them afterwards. This way we would have a clear and super stable contract without experimental stuff.

dbu:
okay, that makes sense to me. with a note that this will become
deprecated and move to the HTTPlug main repo once PSR is released.

sagikazarmark:
Sure. @joelwurtz I've setup `httplug-async` repo. Can you push content there? Keep in mind that it must require the `httpluh` repo because of exceptions.

joelwurtz:
Sure, should i keep the same namespace or put everything under Http\Client\Async ?

sagikazarmark:
I think you should keep it as it is only a temporary solution.

dbu:
yeah, it technically works with composer to have several packages
provide classes/interface in the same namespace

joelwurtz:
done in https://github.com/php-http/httplug-async

dbu:
hurray!

PR #78, opened 2015-10-27
remove docblock about multiple requests handling

xabbuh:
With #75 being merged the `HttpClient` interface no longer supports sending multiple requests at once.


hannesvdvreken:
:+1:

dbu:
ups, that slipped through. thanks

PR #79, opened 2015-11-01
Update PHPSpec to make builds running on PHP 7

sagikazarmark:
Yay, tests are running for PHP 7

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "psr/http-message": "^1.0"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


dbu:
should we take php7 out of allowed failures now?

sagikazarmark:
Not sure. PHPSpec is still alpha. Once it gets stable, we can i think.

joelwurtz:
There should be a stable release in one week :)

sagikazarmark:
Great. After that we can remove the dev tag and remove php7 from allow failures.

PR #80, opened 2015-11-08
A few more improvements

sagikazarmark:
After proofreading the code, I've found a few minor things, corrected them.

I also replaced gitter with slack (which meanst that ):

http://slack.httplug.io


joelwurtz:
LGTM

Issue #86, opened 2015-11-22
Promises


sagikazarmark:
## Implementations

Currently we have our own Promise interface, which is fine. It seems that we won't have a Promise PSR anytime soon.

However, there are already awesome promise packages:
- https://github.com/guzzle/promises
- https://github.com/reactphp/promise/

I remember that Guzzle used to use ReactPHP's promise implementation. Maybe @mtdowling could tell us why they created their own?

Anyway: would it make sense to rely on third party promise implementations instead of ours?

Possible ways:
- Keep the interface in the contract and use wrappers around third-party promises
- Drop our interface and rely on a third-party directly (which would require relying on an implementation in the contract)
## Convenience methods

Currently we have two convenience methods: `getResponse` and `getException`. These are what make our interface a "special" promise implementation. As the above two implementations, ours claim to be an implementation of [Promises/A+](https://promisesaplus.com/) standard. Although I see the point in these methods, I think we should probably get rid of them.

Here are a few reasons:
- If a Promise PSR ever becomes real, we would still need to extend it. In this case we couldn't just accept any kind of promises to be used.
- The same case, if we want to use a third-party implementation now.
- It is not part of the original standard, but as I can see, returning the value in the `wait` method is.

I propose removing those method. However in that case we need a "standard" way of getting the value and deciding whether it is an exception, a response, or an invalid value. Checking the type is obvious, but might led to duplicated code. Maybe we could check how Guzzle does it.

joelwurtz:
I don't mind dropping the getResponse / getException method, problem is more in the return type of the onFulfilled callable or the throw type, that's must be fixed. Having those method makes, IMO, user / implementors understand more what the promise must return / use.

A valid promise specification allow anything for the value or for the reason. However in our case having something different from ResponseInterface / HttpException will break everything. (A plugin can return a string in the onCallable, this would be valid for the promise specification, but will break everything for the pluginChain)

> ours claim to be an implementation of Promises/A+ standard

Not at all, but it may be not clear, it's an **extension** of this standard (basically it's the promise a+ standard with addition that the value and reason has a type: ResponseInterface / Http\Exception)

sagikazarmark:
> Having those method makes (maybe) user / implementors understand more what the promise must return / use.

I would say we should clearly document this. That would probably be enough. 

> However in our case having something different from ResponseInterface / HttpException will break everything.

[...]

PR #90, opened 2015-12-12
Use php-http/promise for the async interface

joelwurtz:
Like discussed in #86


sagikazarmark:
Shouldn't actually be merged until the Promise getResponse/getException methods are removed from everywhere.

joelwurtz:
Yeah but we need this to be merged to upgrade other libs :)

sagikazarmark:
We can rely on the feature branch for the time being, then magically merge everything at the same time.

joelwurtz:
All pr are done IMO

PR #93, opened 2015-12-14
Add exception explanation, fixes #87

### LINKED ISSUE #87: Exceptions
## Exceptions in promises

If I understand it well, in case of async requests the async request method should NEVER EVER throw an exception, but return and reject a promise.

This is currently not the case:
https://github.com/php-http/guzzle6-adapter/blob/master/src/Guzzle6Promise.php#L66
## SPL or not

Although we agreed that SPL exceptions can be used to determine exceptions not closely related to the domain, we might want to reconsider this.

The point of the common Exception interface is that every exception can be caught. On the other side: creating a custom version of every SPL exception which implements our interface seems to be dumb.

The ideal solution would be creating domain exceptions for each case we would use an SPL one.

For example in this case:
https://github.com/php-http/guzzle6-adapter/blob/master/src/Guzzle6Promise.php#L66

This decision should take effect in the Promise interface as well: currently it is restricted to our Exception interface.

**Update:** Using anonymus classes would be a good solution, but that is only available in PHP 7. Maybe we could place custom extensions in one of our utils repo.

sagikazarmark:
@dbu @joelwurtz please help me with the explanation. :smile:

### DIFF
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -21,7 +21,8 @@ interface HttpClient
      *
      * @return ResponseInterface
      *
-     * @throws Exception
+     * @throws \Http\Client\Exception If an error happens during processing the request.
+     * @throws \Exception             If processing the request is impossible (eg. bad configuration).
      */
     public function sendRequest(RequestInterface $request);
 }
--- a/src/HttpAsyncClient.php
+++ b/src/HttpAsyncClient.php
@@ -15,9 +15,13 @@ interface HttpAsyncClient
     /**
      * Sends a PSR-7 request in an asynchronous way.
      *
+     * Exceptions related to processing the request are available from the returned Promise.
+     *
      * @param RequestInterface $request
      *
      * @return Promise
+     *
+     * @throws \Exception If processing the request is impossible (eg. bad configuration).
      */
     public function sendAsyncRequest(RequestInterface $request);
 }


dbu:
looks good to me. for async, we could mention in the text part that exceptions that should be handled like network issues and server problems are available from the promise, as they might not be known yet when the sendAsyncRequest method returns.

sagikazarmark:
Better now?

(Should be squashed)

dbu:
looks good to me, yes!

sagikazarmark:
@joelwurtz ping

src/HttpClient.php
```diff
@@ -21,7 +21,8 @@
      *
      * @return ResponseInterface
      *
-     * @throws Exception
+     * @throws Exception  If an error happens during processing the request.
+     * @throws \Exception If processing the request is impossible (eg. bad configuration).
```
joelwurtz: Can we use this exception naming instead:

``` php
@throws \Http\Exception
@throws \Exception
```

Currently it's feel a bit weird and i'm afraid of misunderstanding from user, if it's read like that

joelwurtz:
Aside from my comment, :+1: for me

src/HttpClient.php
dbu: agreed for this situation. but in general we should keep using the `use` statement for phpdoc as well.

PR #102, opened 2015-12-29
Removed dependency for php-http/promise

Nyholm:
There is no technical dependency on `php-http/promise`. Since we only use it in docs I think we should remove it. 

Packages that need the `php-http/promise` should require it directly. See https://github.com/php-http/guzzle6-adapter/pull/26

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,8 +16,7 @@
     ],
     "require": {
         "php": ">=5.4",
-        "psr/http-message": "^1.0",
-        "php-http/promise": "^0.1"
+        "psr/http-message": "^1.0"
     },
     "require-dev": {
         "phpspec/phpspec": "^2.4",
--- a/src/HttpAsyncClient.php
+++ b/src/HttpAsyncClient.php
@@ -2,11 +2,10 @@
 
 namespace Http\Client;
 
-use Http\Promise\Promise;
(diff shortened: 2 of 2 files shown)


sagikazarmark:
There is: the async client requires it.

Nyholm:
No, it does not as I can see it. 

The async client interface use it in the PHP doc to say that the method should return a Promise. "Technically", it is no requirement.

sagikazarmark:
See the reasoning here: php-http/guzzle6-adapter#26

I think it is still part of the contract, no matter if it is in the same package or not.

In the future if we release new versions of the promise library (for example a major version with BC breaks) we have to make sure that it continues working with the relevant httplug version. With other words: httplug needs to specify the supported versions, since it is part of the contract.

Nyholm:
Hm, yeah, it makes sense. You are correct. 
If this was PHP7 and we were using return types it would be obvious that we should have php-http/promise in composer.json. 

Thank you. 

I stil think https://github.com/php-http/guzzle6-adapter/pull/26 is valid though.

sagikazarmark:
> I stil think php-http/guzzle6-adapter#26 is valid though.

I will have to think about it. I can imagine cases where you might need bigger control over the exact promise version.

PR #107, opened 2016-01-25
Add Promise return type doc in Async client

sagikazarmark:


src/HttpAsyncClient.php
```diff
@@ -19,7 +19,7 @@
      *
      * @param RequestInterface $request
      *
-     * @return Promise
+     * @return Promise MUST return with a PSR-7 Response.
```
joelwurtz: Not a fan :p

More like : 

The promise resolve a ResponseInterface or fail with a Http\Client\Exception

sagikazarmark: Sounds good to me.

sagikazarmark:
Looks better?

joelwurtz:
Would remove "MUST" as the message is more for the user, and people implementing the library must conform with the view of the user

joelwurtz:
A part from that :+1:

PR #109, opened 2016-01-26
Prepare stable release

sagikazarmark:


composer.json
```diff
@@ -34,9 +34,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.0-dev"
+            "dev-master": "1.1-dev"
```
dbu, line 37: my practice so far was to leave this at 1.0 until i introduce the first new features. but its easily forgotten, maybe this way is better. we can still tag 1.0.1 and so on on master as long as we are sure there are no new features.

sagikazarmark, line 37: Exactly. 1.1.x does not mean, that we cannot make 1.0.x versions IMO.

joelwurtz:
:+1:

PR #128, opened 2017-08-18
Adding PSR package

### LINKED ISSUE #51: Make HTTP Client a PSR standard
This is not an immediate thing, but more of a discussion about the future:

[liuggio suggested to propose the client as PSR](https://twitter.com/liuggio/status/645971772261707776). i think this is a viable idea, but will take more time. see also [the discussion on PSR that liuggio referse to](https://groups.google.com/forum/#!topic/php-fig/Ew36Ng5EwXE).

i think at this point we should just release what we have, as it will improve the situation and bring PSR-7 for generic HTTP Client consumers this year.

but we should try to get pinged/involved with the PSR discussion when they pick this topic up again. looking at the interfaces they throw around, i see there is little difference at the core between a server and a HTTP client: you have a request and want a response. whether the implementation produces the response itself or does a web request is more of an implementation detail. though the situation you use one or the other will often look different, and e.g. the kind of errors that can happen look different...

with the time just PSR-7 took, i think it will take a long time to get a PSR for this topic released and we should not wait, rather go with this abstraction and gather experience that can help build the PSR.

/cc @weierophinney, @liuggio

Nyholm:
Yes, This PR is very premature. I just want to make sure we all agree (and look forward to) that this is eventually going to be merged. 

This will fix #51

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Change Log
 
+## 2.0.0 - UNRELEASED
+
+This version contains is technically a BC break because we drop PHP5 and add return 
+type annotation to support `psr/http-client`. 
+
+### Added
+
+- Support for PSR-18 (Http client).
+
+### Removed
+
+- PHP 5 support
+
+### Changed
+
+- [BC Break] `HttpClient::sendRequest(RequestInterface $request)` has a return type annotation. The mew 
+signature is `HttpClient::sendRequest(RequestInterface $request): ResponseInterface`.  
+- [BC Break] `RequestException::getRequest()` has a return type annotation. The new 
+signature is `RequestException::getRequest(): RequestInterface`.   
 
 ## 1.1.0 - 2016-08-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,12 @@
 
 ## 2.0.0 - UNRELEASED
 
-This version contains is technically a BC break because we drop PHP5 and add return 
-type annotation to support `psr/http-client`. 
+This version is no BC break for consumers using HTTPlug. However, HTTP clients that
+implement HTTPlug need to adjust because we add return type declarations.
 
 ### Added
 
-- Support for PSR-18 (Http client).
+- Support for PSR-18 (HTTP client).
 
 ### Removed
 
@@ -15,7 +15,7 @@ type annotation to support `psr/http-client`.
 
 ### Changed
 
-- [BC Break] `HttpClient::sendRequest(RequestInterface $request)` has a return type annotation. The mew 
+- [BC Break] `HttpClient::sendRequest(RequestInterface $request)` has a return type annotation. The new 
 signature is `HttpClient::sendRequest(RequestInterface $request): ResponseInterface`.  
 - [BC Break] `RequestException::getRequest()` has a return type annotation. The new 
 signature is `RequestException::getRequest(): RequestInterface`.   
--- a/src/Exception/HttpException.php
+++ b/src/Exception/HttpException.php
@@ -37,7 +37,12 @@ class HttpException extends RequestException
         $this->code = $response->getStatusCode();
     }
 
-    public function getResponse(): ResponseInterface
+    /**
+     * Returns the response.
+     *
+     * @return ResponseInterface
+     */
+    public function getResponse()
     {
         return $this->response;
     }
@@ -49,7 +54,7 @@ class HttpException extends RequestException
         RequestInterface $request,
         ResponseInterface $response,
         \Exception $previous = null
-    ): self {
+    ) {
         $message = sprintf(
             '[url] %s [http method] %s [status code] %s [reason phrase] %s',
             $request->getRequestTarget(),
--- a/src/Exception/HttpException.php
+++ b/src/Exception/HttpException.php
@@ -37,11 +37,6 @@ class HttpException extends RequestException
         $this->code = $response->getStatusCode();
     }
 
-    /**
-     * Returns the response.
-     *
-     * @return ResponseInterface
-     */
     public function getResponse(): ResponseInterface
     {
         return $this->response;
@@ -49,12 +44,6 @@ class HttpException extends RequestException
 
     /**
      * Factory method to create a new exception with a normalized error message.
-     *
-     * @param RequestInterface  $request
-     * @param ResponseInterface $response
-     * @param \Exception|null   $previous
-     *
-     * @return HttpException
      */
     public static function create(
         RequestInterface $request,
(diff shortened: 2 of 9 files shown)


src/HttpAsyncClient.php
```diff
@@ -23,5 +23,5 @@
      *
      * @throws \Exception If processing the request is impossible (eg. bad configuration).
      */
-    public function sendAsyncRequest(RequestInterface $request);
+    public function sendAsyncRequest(RequestInterface $request): Promise;
```
Nyholm: Is this problem when we are doing the same update for the Async HTTP client PSR?

dbu: there needs to be a new major version anyways when a promise PSR is out, so we might as well do this change now to make it more explicit.

src/Exception/HttpException.php
```diff
@@ -42,7 +42,7 @@ public function __construct(
      *
      * @return ResponseInterface
```
dbu: in our projects we started the rule that when `@return` just repeats the return type declaration and has no valuable information, we remove the `@return` line. 

(in that vein "Returns the response" is redundant to the method name. either there is more details to explain or there should be no comment...)

Nyholm: I agree. 👍 

I will remove these redundant comments.

src/HttpAsyncClient.php
joelwurtz: I would like to be more permissive in the async implementation, to allow every value and exception in our promise, this would be better, after years of use it was a mistake to reduce the scope IMO.

Nyholm: Fair, so what do you suggest? I've written `Promise` here, not `HttpPromise`.

joelwurtz: In fact, it's in another package, we would need to update the promise lib and then allow 2.0 here, but maybe it will not be bc break for this package, so nothing to change here

joelwurtz: (We would still need to release a 2.1 with the package dependency upgrade to allow 1.0 || 2.0 for the promise lib)

joelwurtz: In fact what do you think of using https://amphp.org/amp/promises/ interface ? there is an adapter for react and we could drop support of async for guzzle (as it's not really async in fact, it's only parallel)

src/Exception/HttpException.php
```diff
@@ -37,30 +37,19 @@ public function __construct(
         $this->code = $response->getStatusCode();
     }
 
-    /**
-     * Returns the response.
-     *
-     * @return ResponseInterface
-     */
-    public function getResponse()
+    public function getResponse(): ResponseInterface
```
Nyholm: This is a BC break: https://3v4l.org/LOOiv

Is this something we want to avoid in this update to make it as smooth as possible?

src/Exception/RequestException.php
```diff
@@ -31,12 +31,7 @@ public function __construct($message, RequestInterface $request, \Exception $pre
         parent::__construct($message, 0, $previous);
     }
 
-    /**
-     * Returns the request.
-     *
-     * @return RequestInterface
-     */
-    public function getRequest()
+    public function getRequest(): RequestInterface
```
Nyholm: Same here

src/Exception/HttpException.php
```diff
@@ -37,30 +37,19 @@ public function __construct(
...
-     * @param \Exception|null   $previous
-     *
-     * @return HttpException
      */
     public static function create(
         RequestInterface $request,
         ResponseInterface $response,
         \Exception $previous = null
-    ) {
+    ): self {
```
Nyholm: Same here

src/HttpAsyncClient.php
Nyholm: Hm.. Im not sure at all...

On one hand: Do as little change as possible. That will make it easier for everyone. 
On the other hand: Can we make the "correct" decision now so there is no need for v3 when we got a PSR for Promise...? Probably not, right?

dbu: how soon is a PSR Promise to be expected? i would assume that will take a couple of years, so if its a clear and good upgrade, i would be ok to have version 2 do something different. then again, to make the PSR move for sync http client as smooth as possible, its maybe better to not limit ourselves here and avoid a BC break

Nyholm:
Im done with this PR. I suggest merging it now and start prepare for a 2.0 release. 

Releasing 2.0 is obviously blocked by a stable release of psr/http-client

Nyholm:
I removed HHVM because it failed because of a PHPSpec error (I think) https://travis-ci.org/php-http/httplug/jobs/409431649

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Change Log
 
+## 2.0.0 - UNRELEASED
+
+This version contains is technically a BC break because we drop PHP5 and add return
```
dbu: <preaching mode on>dropping PHP5 support is no BC break. it just prevents people on an obsolete stack from updating</off>

how do you mean "technically"? i think we should say: "This version is no BC break for consumers *using* HTTPlug. However, HTTP clients that *implement* HTTPlug need to adjust because we add return type declarations."

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Change Log
 
+## 2.0.0 - UNRELEASED
+
+This version contains is technically a BC break because we drop PHP5 and add return 
+type annotation to support `psr/http-client`. 
+
+### Added
+
+- Support for PSR-18 (Http client).
```
dbu: Http or HTTP? (i don't really care but we should be consistent in our doc and consistent with how fig documents write it)

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+
+### Removed
+
+- PHP 5 support
+
+### Changed
+
+- [BC Break] `HttpClient::sendRequest(RequestInterface $request)` has a return type annotation. The mew
```
dbu: s/mew/new/ ;-)

dbu (changes requested):
this looks great! so indeed the transition path looks quite sane.

as for merging, i guess as long as we don't release a version, we should be fine even if it turns out that PSR-18 still changes somehow.

please adjust the branch-alias in composer.json to 2.0-dev

Nyholm:
Thank you for the review. I've corrected the PR accordingly.

dbu (approved):
great, happy with it like this!

i'll wait a bit if @sagikazarmark has input as well

sagikazarmark:
Looks good to me. 👍

sagikazarmark (approved):

Nyholm:
I've changed base to a new 2.x branch.

Nyholm:
Thank you for the reviews.

PR #136, opened 2018-08-03
comment unused dependency

atefBB:
Deprecations: yes.
This dependency should be removed (I think).

### DIFF
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -3,7 +3,7 @@
 namespace Http\Client;
 
 use Psr\Http\Message\RequestInterface;
-use Psr\Http\Message\ResponseInterface;
+// use Psr\Http\Message\ResponseInterface;
 
 /**
  * Sends a PSR-7 Request and returns a PSR-7 response.
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -3,7 +3,6 @@
 namespace Http\Client;
 
 use Psr\Http\Message\RequestInterface;
-// use Psr\Http\Message\ResponseInterface;
 
 /**
  * Sends a PSR-7 Request and returns a PSR-7 response.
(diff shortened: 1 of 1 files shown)


Nyholm:
Hey. 
Thank you for this PR. The import is used here in the comment: https://github.com/php-http/httplug/pull/136/files#diff-add00846710b59ac0af2a07e1ede2747R22

atefBB:
Hey @Nyholm!
What about this: `@return Psr\Http\Message\ResponseInterface`

GrahamCampbell:
> What about this: @return Psr\Http\Message\ResponseInterface

Actually, you'll need a leading slash on that, in order for it to resolve from the global namespace.

Nyholm:
Im not sure what we win by this change... Why is one better than the other?

src/HttpClient.php
```diff
@@ -19,7 +18,7 @@
      *
      * @param RequestInterface $request
      *
-     * @return ResponseInterface
+     * @return Psr\Http\Message\ResponseInterface
```
dbu, line 21: a return type annotation is also a dependency on the class from the other namespace. imho we should use `use` statements for those as well.

sagikazarmark:
I agree with the others, not sure what's the advantage, I'm happy with the current situation. It's a trivial thing anyway. Thanks for the contribution though.

src/HttpClient.php
atefBB: Thanks for the info.

atefBB:
@sagikazarmark @dbu @Nyholm @GrahamCampbell Thank you all!

dbu:
thank you @atefBB for looking at the code and sending a pull request. don't let this discourage you and keep sending pull requests if you see things that you think can be improved! (before doing larger changes, best open an issue to discuss first, to avoid wasting time on the code)

PR #140, opened 2018-09-05
2.x branch alias

Nyholm:
Better alias per https://getcomposer.org/doc/articles/aliases.md

We now allow people to `composer req php-http/httplug:2.0.*`


fbourigault (approved):

Nyholm:
Thank you

PR #144, opened 2018-10-17
Update minimum PHP version to ^7.0 and modify Http\Client\Exception to extend \Throwable

### LINKED ISSUE #142: phpstan complains about Http\Client\Exception
| Q            | A
| ------------ | ---
| Bug?         | maybe
| New Feature? | no
| Version      | 1.1.0

#### Actual Behavior

When building an HTTP client that implements `Http\Client\HttpClient`, I am implementing the `sendRequest()` method, as required by the interface. I have added the `@throws` docblock annotation with `Http\Client\Exception`, since that is what the interface defines. When I run `phpstan`, I receive the following error because `phpstan` detects that the `Exception` interface does not inherit from `Throwable`.

```
$ composer run phpstan
> phpstan analyse MyClient.php --level max --no-progress
 ------ ----------------------------------------------------------------------
  Line   MyClient.php
 ------ ----------------------------------------------------------------------
  32     PHPDoc tag @throws with type Http\Client\Exception is not subtype of
         Throwable
 ------ ----------------------------------------------------------------------

 [ERROR] Found 1 error
```

#### Expected Behavior

I expect static analysis tools to pass without error when defining `@throws \Http\Client\Exception` on a method docblock.

#### Steps to Reproduce

1. Use the `composer.json` and `MyClient.php` files provided below.
2. Run `composer install`
3. Run `composer run phpstan`

##### composer.json

``` json
{
    "require": {
        "php-http/httplug": "^1.1",
        "phpstan/phpstan": "^0.10.3"
    },
    "autoload": {
        "psr-4": {
            "Ramsey\\Example\\": "."
        }
    },
    "scripts": {
        "phpstan": "phpstan analyse MyClient.php --level max --no-progress"
    }
}
```

##### MyClient.php

``` php
<?php

declare(strict_types=1);

namespace Ramsey\Example;

use Http\Client\Exception;
use Http\Client\HttpClient;
use Psr\Http\Message\RequestInterface;
use Psr\Http\Message\ResponseInterface;

class MyClient implements HttpClient
{
    /**
     * @var HttpClient
     */
    private $client;

    /**
     * @param HttpClient $client
     */
    public function __construct(HttpClient $client)
    {
        $this->client = $client;
    }

    /**
     * @return ResponseInterface
     * @throws Exception
     */
    public function sendRequest(RequestInterface $request): ResponseInterface
    {
        return $this->client->sendRequest($request);
    }
}
```

#### Possible Solutions

The `Http\Client\Exception` interface should extend `Throwable`, which would cause `phpstan` and other static analysis tools to run without errors.

I understand this may not be possible, since this library supports versions of PHP earlier than 7.0. This is also why I put "maybe" as an answer to whether this is a bug.

ramsey:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #142 
| License         | MIT

#### What's in this PR?

Updates `Http\Client\Exception` to extend from `\Throwable`. Also updates the library to support only PHP version 7.0.0 and greater, to allow use of the `\Throwable` interface, introduced in PHP 7.

#### Why?

Static analysis tools are complaining when creating HTTP clients that implement `Http\Client\HttpClient`

#### Example Usage

n/a

#### Checklist

*Will update the CHANGELOG if this approach is acceptable to the maintainers.*

#### To Do

n/a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ cache:
         - $HOME/.composer/cache/files
 
 php:
-    - 5.4
-    - 5.5
-    - 5.6
     - 7.0
     - 7.1
     - 7.2
@@ -25,13 +22,11 @@ branches:
 matrix:
     fast_finish: true
     include:
-        - php: 5.4
+        - php: 7.0
           env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" COVERAGE=true TEST_COMMAND="composer test-ci"
-        - php: hhvm
-          dist: trusty
 
 before_install:
-    - if [[ $COVERAGE != true && $TRAVIS_PHP_VERSION != hhvm ]]; then phpenv config-rm xdebug.ini || true; fi
+    - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
     - travis_retry composer self-update
 
 install:
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4",
+        "php": "^7.0",
         "psr/http-message": "^1.0",
         "php-http/promise": "^1.0"
     },
--- a/src/Exception.php
+++ b/src/Exception.php
@@ -7,6 +7,6 @@ namespace Http\Client;
  *
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
-interface Exception
+interface Exception extends \Throwable
 {
 }


sagikazarmark:
I believe this should happen as part of the PSR-18 upgrade, until then I see no point in this change.

/cc @Nyholm

Nyholm:
Yeah, check out the 2.x branch. I think we should merge that one to master now.

Nyholm:
This is fixed in master. Thank you for this PR.

PR #150, opened 2018-10-31
tweak readme

dbu:
- improve wording in the readme.
- use 2.x-dev as branch alias so its still correct when we move to 2.1 and so on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,15 +15,26 @@
 
 ## Intro
 
-HTTPlug is the predecessor of [PSR-18](http://www.php-fig.org/psr/psr-18/)
-HTTP Client standard built on [PSR-7](http://www.php-fig.org/psr/psr-7/) HTTP messages.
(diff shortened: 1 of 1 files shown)


sagikazarmark (approved):
Thanks!

composer.json
```diff
@@ -35,7 +35,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0.x-dev"
```
GrahamCampbell: Including the minor version is better, otherwise compose will resolve things like 2.55.x to master, when in fact master is 2.0.x series.

GrahamCampbell (changes requested):

composer.json
dbu: i was afraid we will forget to up this to 2.1.x-dev once we start that version. but will revert this change then.

PR #151, opened 2019-01-21
Remove exception docblock

gmponos:
I think according to PSR this exception should not be thrown... :/

### DIFF
--- a/src/HttpClient.php
+++ b/src/HttpClient.php
@@ -3,22 +3,13 @@
 namespace Http\Client;
 
 use Psr\Http\Client\ClientInterface;
-use Psr\Http\Message\RequestInterface;
-use Psr\Http\Message\ResponseInterface;
 
 /**
- * Sends a PSR-7 Request and returns a PSR-7 response.
+ * {@inheritdoc}
  *
- * @author GeLo <geloen.eric@gmail.com>
- * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
- * @author David Buchmann <mail@davidbu.ch>
+ * Provide the Httplug HttpClient interface for BC.
+ * You should typehint Psr\Http\Client\ClientInterface in new code
  */
 interface HttpClient extends ClientInterface
 {
-    /**
-     * Sends a PSR-7 request.
-     *
-     * @throws \Http\Client\Exception If an error happens during processing the request.
-     */
-    public function sendRequest(RequestInterface $request): ResponseInterface;
 }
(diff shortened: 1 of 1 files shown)


dbu (commented):
correct. looking at the psr-18 ClientInterface that we extend here, i think we should actually drop all code in here and make this
```
/**
 * {@inheritdoc}
 * 
 * Provide the Httplug HttpClient interface for BC. You should typehint Psr\Http\Client\ClientInterface in new code
 */
interface HttpClient extends ClientInterface
{
}
```

the method definition is a copy of the psr method and therefore does nothing. do you want to adjust this accordingly?

gmponos:
>do you want to adjust this accordingly?

Done.. check again :)

dbu (approved):
@Nyholm am i missing something or is this what the interface should look like now?

Nyholm:
Nope, We still throw HTTPlug exceptions. The exceptions however extends PSR18. 

Im 👎

gmponos:
Hmm.. I missed that.. @Nyholm  is correct... So shall I revert [here](https://github.com/php-http/httplug/pull/151/commits/4d0b8a5a75ca20de12a164f75a0a3313a365360d) before @dbu comment...

are you all good with this?

Nyholm:
To remove `\Exception` and update the description on the other line makes sense, yes.

dbu:
> We still throw HTTPlug exceptions. The exceptions however extends PSR18.

is there any value in explicitly stating that, however? imho people should be using the PSR-18 exception interfaces, not the ones from the HTTPlug BC layer. as the exceptions extend the PSR-18 interface, the phpdoc of the base interface is not wrong.

Nyholm:
If you are using a HTTPlug 2.0 client you may catch both HTTPlug exceptions and PSR18 exceptions.

>  imho people should be using the PSR-18 exception interfaces
Sure, and they may.

dbu:
@Nyholm so you would say we should add the method signature again to add the phpdoc for the httplug exceptions?

joelwurtz:
> If you are using a HTTPlug 2.0 client you may catch both HTTPlug exceptions and PSR18 exceptions.

I don't agree with that, catching an HTTPlug is now a detail of implementation in 2.0, and we should not spec this, as this will allow people to correctly depend on PSR 18 and completly remove our interface latter.

dbu:
@Nyholm are you okay to merge with this or do you disagree with the last comments from joel and me?

PR #153, opened 2019-10-02
Use static instead of self for instancing HttpException

### LINKED ISSUE #152: HttpException create method self vs static
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? |  yes
| Version      | 1.1.0

#### Actual Behavior

method create returns HttpException instance

#### Expected Behavior

Should return an instance of calling class eg ServerErrorException::create() should return 
ServerErrorException

#### Steps to Reproduce

call ServerErrorException::create()

#### Possible Solutions

Change self to static keyword in create method of HttpException class

kpn13:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #152
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

Use `new static()` instead of `new self()` for instancing `Http\Client\Exception\HttpException` to really instanciate the final class which called the `create()` method and not just the parent class.

#### Checklist

### DIFF
--- a/src/Exception/HttpException.php
+++ b/src/Exception/HttpException.php
@@ -63,6 +63,6 @@ class HttpException extends RequestException
             $response->getReasonPhrase()
         );
 
-        return new self($message, $request, $response, $previous);
+        return new static($message, $request, $response, $previous);
     }
 }


dbu (approved):
thanks!

PR #156, opened 2019-12-14
Test on 7.4

Big-Shark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | No
| License         | MIT

PHP 7.4 released, and we need run tests for 7.4 too


dbu (approved):
thank you!

while it does not end up in an error, there is a warning in the 7.4 build because of https://github.com/phpspec/phpspec/issues/1274

we seem to be on phpspec 2.4 atm, while the release that fixes the warning is phpspec 6.1 :scream: 

i will merge as is, as its only a warning about the testing toolchain. if you want to try to upgrade the project to use phpspec 6.1 that would be awesome. (you'd need to allow some older phpspec version still, if we can support both, as 6.1 is only for php 7.2 and newer)

PR #157, opened 2019-12-16
Update phpspec

Big-Shark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | phpspec/phpspec#1274
| Documentation   | no
| License         | MIT

#### What's in this PR?

Just update phpspec version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "php-http/promise": "^1.0"
     },
     "require-dev": {
-        "phpspec/phpspec": "^4.3|^5.0|^6.0",
+        "phpspec/phpspec": "^4.3.4|^5.0|^6.0",
         "friends-of-phpspec/phpspec-code-coverage" : "^4.1"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "php-http/promise": "^1.0"
     },
     "require-dev": {
-        "phpspec/phpspec": "^2.4",
+        "phpspec/phpspec": "^4.0|^5.0|^6.0",
         "henrikbjorn/phpspec-code-coverage" : "^1.0"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     },
     "require-dev": {
         "phpspec/phpspec": "^4.0|^5.0|^6.0",
-        "henrikbjorn/phpspec-code-coverage" : "^1.0"
+        "friends-of-phpspec/phpspec-code-coverage" : "^4.1"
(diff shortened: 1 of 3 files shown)


dbu:
thanks! looking at the lowest version build failure, i think we need a higher minimum version of either the coverage thing or phpspec itself

composer.json
```diff
@@ -21,8 +21,8 @@
         "php-http/promise": "^1.0"
     },
     "require-dev": {
-        "phpspec/phpspec": "^2.4",
-        "henrikbjorn/phpspec-code-coverage" : "^1.0"
+        "phpspec/phpspec": "^4.0|^5.0|^6.0",
```
GrahamCampbell: phpspec 4 can be dropped?

GrahamCampbell (changes requested):

composer.json
Big-Shark: @GrahamCampbell In the current moment httplug support php version 7.0, but phpspec 5.0 support only 7.1, and I added phpspec 4.0 for php 7.0

GrahamCampbell: Oh, for some reason I though this package was PHP 7.1+.

In which case, we should change it to `^4.2` probably?

Big-Shark: @GrahamCampbell Why not ^4.3?

dbu: for testing tools, i am happy with the most strict restrictions that still allow to build, if ^4.3 works thats better than ^4.2. we don't give anybody more flexibility if we allow 4.2

dbu:
can you please try phpstan 4.3.4? seems 4.3.0 still has that configuration bug

composer.json
```diff
@@ -21,8 +21,8 @@
         "php-http/promise": "^1.0"
     },
     "require-dev": {
-        "phpspec/phpspec": "^2.4",
-        "henrikbjorn/phpspec-code-coverage" : "^1.0"
+        "phpspec/phpspec": "^4.3|^5.0|^6.0",
+        "friends-of-phpspec/phpspec-code-coverage" : "^4.1"
```
dbu, line 25: use ^4.1.2 of this, to avoid the lowest-version failure?

Big-Shark:
@dbu This is strange, this is not best solution, but this is work.

dbu (approved):
oh, i did not realize we had coverage on the lowest version build only... yeah i agree with your solution, don't see why we would want coverage on lowest version build.

dbu:
thanks a lot!

PR #158, opened 2019-12-16
PSR-18: Network / Request exception inheritance

### LINKED ISSUE #155: PSR-18: Network / Request exception inheritance
| Q            | A
| ------------ | ---
| Bug?         | possibly
| New Feature? | no
| Version      | `2.0.0`

This could be the intended behaviour, so please close if so.

#### Actual Behavior

[NetworkException](https://github.com/php-http/httplug/blob/v2.0.0/src/Exception/NetworkException.php) extends [RequestException](https://github.com/php-http/httplug/blob/v2.0.0/src/Exception/RequestException.php)

#### Expected Behavior

According to https://www.php-fig.org/psr/psr-18/meta/ I wouldn't expect _NetworkException_ to extend _RequestException_

>The domain exceptions NetworkExceptionInterface and RequestExceptionInterface define a contract very similar to each other. The chosen approach is to not let them extend each other because inheritance does not make sense in the domain model. A RequestExceptionInterface is simply not a NetworkExceptionInterface

#### Steps to Reproduce

```php
try {
    throw new \Http\Client\Exception\NetworkException(
        'some network problem',
        $someRequest
    );
} catch (\Psr\Http\Client\RequestExceptionInterface $e) {
    // Exception is caught here. 
    echo 'caught request exception';
} catch (\Psr\Http\Client\NetworkExceptionInterface $e) {
    // Expect to catch here.
    echo 'caught network exception';
}
```

#### Possible Solutions

- Change _NetworkException_ so that it doesn't extend _RequestException_, then do one of the following:
  - Implement `::getRequest()` in each _(or use a trait)_
  - Extend a class such as "RequestAwareException" that provides `::getRequest()` method
- Document the current structure (instruct to catch `\Psr\Http\Client\NetworkExceptionInterface` before `\Psr\Http\Client\RequestExceptionInterface`)

davidgrayston:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?   | no
| BC breaks?      | changed exception interface implementation
| Deprecations?   | no
| Related tickets | fixes #155 

#### What's in this PR?

- `Http\Client\Exception\NetworkException` no longer extends `Http\Client\Exception\RequestException`
- `Http\Client\Exception\NetworkException` now extends `Http\Client\Exception\TransferException`
- `Http\Client\Exception\NetworkException` and `Http\Client\Exception\RequestException` both use `Http\Client\Exception\RequestAwareTrait` (not sure if this is the best approach?)

#### Example Usage

``` php
try {
    throw new \Http\Client\Exception\NetworkException(
        'some network problem',
        $someRequest
    );
} catch (\Psr\Http\Client\RequestExceptionInterface $e) {
    echo 'caught request exception';
} catch (\Psr\Http\Client\NetworkExceptionInterface $e) {
    echo 'caught network exception';
}
```

#### Checklist

- [X] Updated CHANGELOG.md to describe bugfix

### DIFF
--- a/src/Exception/RequestAwareTrait.php
+++ b/src/Exception/RequestAwareTrait.php
@@ -7,7 +7,7 @@ use Psr\Http\Message\RequestInterface;
 trait RequestAwareTrait
 {
     /**
-     * @var \Psr\Http\Message\RequestInterface
+     * @var RequestInterface
      */
     private $request;
 
--- a/src/Exception/RequestAwareTrait.php
+++ b/src/Exception/RequestAwareTrait.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Http\Client\Exception;
+
+use Psr\Http\Message\RequestInterface;
+
+trait RequestAwareTrait
+{
+    /**
+     * @var \Psr\Http\Message\RequestInterface
+     */
+    private $request;
+
+    /**
+     * @param \Psr\Http\Message\RequestInterface $request
+     */
+    private function setRequest(RequestInterface $request)
+    {
+        $this->request = $request;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getRequest(): RequestInterface
+    {
+        return $this->request;
+    }
+}
--- a/src/Exception/RequestAwareTrait.php
+++ b/src/Exception/RequestAwareTrait.php
@@ -11,9 +11,6 @@ trait RequestAwareTrait
      */
     private $request;
 
-    /**
-     * @param \Psr\Http\Message\RequestInterface $request
-     */
     private function setRequest(RequestInterface $request)
     {
         $this->request = $request;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- `Http\Client\Exception\NetworkException` no longer extends `Http\Client\Exception\RequestException`,
+  in accordance with [PSR-18](https://www.php-fig.org/psr/psr-18/)
 
 ## [2.0.0] - 2018-10-31
 
--- a/spec/Exception/NetworkExceptionSpec.php
+++ b/spec/Exception/NetworkExceptionSpec.php
@@ -17,8 +17,28 @@ class NetworkExceptionSpec extends ObjectBehavior
         $this->shouldHaveType('Http\Client\Exception\NetworkException');
     }
 
-    function it_is_a_request_exception()
+    function it_is_a_transfer_exception()
     {
-        $this->shouldHaveType('Http\Client\Exception\RequestException');
+        $this->shouldHaveType('Http\Client\Exception\TransferException');
+    }
+
+    function it_implements_psr_network_exception_interface()
+    {
+        $this->shouldHaveType('Psr\Http\Client\NetworkExceptionInterface');
+    }
+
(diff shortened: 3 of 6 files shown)


src/Exception/RequestAwareTrait.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Http\Client\Exception;
+
+use Psr\Http\Message\RequestInterface;
+
+trait RequestAwareTrait
+{
+    /**
+     * @var \Psr\Http\Message\RequestInterface
```
dbu: annotation can also profit from the `use` statement

src/Exception/RequestAwareTrait.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    private $request;
+
+    /**
+     * @param \Psr\Http\Message\RequestInterface $request
```
dbu: please remove this annotation, the parameter type declaration is enough.

dbu (approved):
thanks! can you do the 2 phpdoc tweaks please?

src/Exception/RequestAwareTrait.php
davidgrayston: Apologies, I don't think I understand... do we want a `@uses \Http\Client\Exception\RequestAwareTrait` on each of the exceptions or a `@used-by ::setRequest()` and `@used-by ::getRequest()` on this private `$request` property? (or something else?)

davidgrayston: Ah, this doesn't need to be a fully qualified namespace, that makes sense - will update

Nyholm:
Could we really merge this? Isnt this considered a BC break. 

Also, the current implementation is not a violation on PSR-18 if I recall correctly.

dbu:
@Nyholm did you see the motivation in #155 for this? i think it is indeed quite unexpected that catching the RequestExceptionInterface will hide NetworkExceptionInterface and you would need to order the catches inverse if you want to separate them. both exceptions still extend ClientExceptionInterface, which would be the thing to catch/typehint if you mean both exceptions.

i agree that this might trip up somebody in some way, but consider it a design bug with potential for BC breaks if users have relied on the bug. there is however a real risk of psr-18 consuming applications not working as expected when using httplug as it is now, with the mixed-up exceptions. my suggestion is to do a new minor version rather than a patch release and document it in the changelog. do you think that is not strict enough?

GrahamCampbell:
Well, all bug fixes are BC breaks by their very nature. Usually a breaking change is defined as a BC break that is not  bug fix. :)

sagikazarmark:
Well, this is kind of an interesting "incompatibility" between the HTTPlug and PSR design ideas.

Ideally, there should always be at least a `catch` for `Http\Client\Exception` as well, so I would assume exceptions will be caught one way or another.

Those who specifically want to handle `NetworkExceptions` will not be affected. Those who just want to catch all exceptions should use the interface instead.

So I think we can classify this as a bug and fix it accordingly.

dbu:
> Ideally, there should always be at least a catch for Http\Client\Exception as well, so I would assume exceptions will be caught one way or another.

imho the correct thing for PSR users to catch all client exceptions is to catch `Psr\Http\Client\RequestExceptionInterface` and not the implementation exception. but yeah, there is a common base thing to catch all kind of exceptions.

dbu:
@Nyholm okay if i merge this for a 2.1 release, or do you still object?

dbu:
haha, of course this broke the httplug library that i maintain: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/474 :-/  (because i misunderstood the exceptions of httplug when writing that code)

PR #169, opened 2022-02-17
remove slack badge as it does not work

dbu:
@sagikazarmark did you set this up? https://slack.httplug.io/ seems to not work. do you want to fix it or should we remove the badge?


dbu:
i merged it. we can revert the merge if the badge can be fixed.

PR #171, opened 2023-04-11
Deprecate Http\Client\HttpClient and allow psr/http-message v2

nicolas-grekas:
| Q               | A
| --------------- | ---
| Bug fix?         | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | -
| Documentation   | TBD
| License         | MIT

Let's be more explicit about the move to PSR-18.

I'm not throwing a runtime deprecation notice because that might be too annoying for the existing clients, especially those in the `php-http/*` namespace.


composer.json
```diff
@@ -22,11 +22,11 @@
         "php": "^7.1 || ^8.0",
         "php-http/promise": "^1.1",
         "psr/http-client": "^1.0",
-        "psr/http-message": "^1.0"
+        "psr/http-message": "^1.0 || ^2.0"
```
dbu, line 25: are the test failures related to installing with 2.0? the lowest builds are all green.

src/HttpClient.php
```diff
@@ -9,6 +9,8 @@
  *
  * Provide the Httplug HttpClient interface for BC.
  * You should typehint Psr\Http\Client\ClientInterface in new code
+ *
+ * @deprecated since version 2.4, use Psr\Http\Client\ClientInterface instead.
```
dbu:
```suggestion
 * @deprecated since version 2.4, Psr\Http\Client\ClientInterface instead. See https://www.php-fig.org/psr/psr-18/ 
```

CHANGELOG.md
```diff
@@ -7,7 +7,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-## [Unreleased]
+## [2.4.0] - Unreleased
+
+### Changed
+
+- Allow `psr/http-message` v2 in addition to v1
+- Deprecate `Http\Client\HttpClient`, use `Psr\Http\Client\ClientInterface` instead
```
dbu:
```suggestion
- Deprecate `Http\Client\HttpClient`, use `Psr\Http\Client\ClientInterface` from [PSR-18](https://www.php-fig.org/psr/psr-18/) instead
```

nicolas-grekas:
Sibling PR at https://github.com/php-http/message-factory/pull/42

composer.json
nicolas-grekas: fixed, all green!

src/HttpClient.php
nicolas-grekas: updated

PR #175, opened 2024-03-15
Make nullable parameter types explicit

derrabus:
#### What's in this PR?

This PR turns implicit nullable parameter types to explicit ones.

#### Why?

PHP allows to implicitly declare a nullable parameter type by setting the parameter's default value to `null`. However, in PHP 8.4 this syntax has been deprecated.

https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

see also php-http/promise#34

### DIFF
--- a/src/Promise/HttpFulfilledPromise.php
+++ b/src/Promise/HttpFulfilledPromise.php
@@ -18,7 +18,7 @@ final class HttpFulfilledPromise implements Promise
         $this->response = $response;
     }
 
-    public function then(callable $onFulfilled = null, callable $onRejected = null)
+    public function then(?callable $onFulfilled = null, ?callable $onRejected = null)
     {
         if (null === $onFulfilled) {
             return $this;
--- a/src/Exception/HttpException.php
+++ b/src/Exception/HttpException.php
@@ -26,7 +26,7 @@ class HttpException extends RequestException
         $message,
         RequestInterface $request,
         ResponseInterface $response,
-        \Exception $previous = null
+        ?\Exception $previous = null
     ) {
         parent::__construct($message, $request, $previous);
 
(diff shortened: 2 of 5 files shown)


src/Promise/HttpFulfilledPromise.php
```diff
@@ -18,7 +18,7 @@ public function __construct(ResponseInterface $response)
         $this->response = $response;
     }
 
-    public function then(callable $onFulfilled = null, callable $onRejected = null)
+    public function then(?callable $onFulfilled = null, ?callable $onRejected = null)
```
dbu, line 21: can we do this without the interface changing? if not, this bumps the minimum requirement of promise to the version that will contains https://github.com/php-http/promise/pull/34

dbu (commented):
thanks. i don't expect these classes to be extended, and the constructors certainly are no issue. so we don't need to consider this a BC break.

src/Promise/HttpFulfilledPromise.php
derrabus: > can we do this without the interface changing?

Yes, this is a purely syntactical change.

derrabus:
> we don't need to consider this a BC break.

It's not, even if people extended the classes. This change doesn't break anything.

dbu:
thanks a lot!

derrabus:
Thank you!

eiriksm:
Thanks a lot for the work here. Would it be possible to cut a new release with this fix included? :nerd_face: 

Thanks again for your work on this package!

dbu:
sure: https://github.com/php-http/httplug/releases/tag/2.4.1

there are currently merge requests on other php-http repositories to do the same cleanup

PR #176, opened 2024-09-23
Run CI on PHP 8.3 and 8.4

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?         | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #175
| Documentation   | -
| License         | MIT

#### What's in this PR?

Run the CI on both PP 8.3 and 8.4.

#### Why?

Validates compatibility with those PHP versions.

#### Example Usage

-

#### Checklist

- [ ] Documentation pull request created (if not simply a bugfix)

#### To Do

- [ ] If the PR is not complete but you want to discuss the approach, list what remains to be done here

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,11 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: ['7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2']
+        php: ['7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4']
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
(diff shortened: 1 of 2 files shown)


.php-cs-fixer.dist.php
```diff
@@ -9,6 +9,7 @@
     ->setRiskyAllowed(true)
     ->setRules([
         '@Symfony' => true,
+        'trailing_comma_in_multiline' => false,
```
dbu, line 12: i would prefer to fix this rather than diverge from the symfony ruleset.

but as you mentioned in https://github.com/php-http/client-common/pull/247 it would make sense to pin the cs fixer to a specific version to avoid random unrelated cs failures in merge requests.

.github/workflows/ci.yml
```diff
@@ -25,6 +25,10 @@ jobs:
           tools: composer:v2
           coverage: none
 
+      - name: Emulate PHP 8.3
+        run: composer config platform.php 8.3.999
```
dbu, line 29: why do we need this? do we have dependencies that limit to php 8.3 and don't install with 8.4?

dbu (commented):
thanks!

.php-cs-fixer.dist.php
GrahamCampbell: We have to do this because the syntax is not compatible with PHP 7.

.github/workflows/ci.yml
GrahamCampbell: phpspec disallows PHP 8.4

dbu (approved):
thanks!

